Issue #4, opened 2022-08-21
Local Cluster integration test


mzitnik:
Our goal is to create a ClickHouse local cluster for Integration tests.

- Spin Kafka Confluent Platform / Amazon MSK
- Spin ClickHouse Cluster
- Tests
   - single topic
   - multitopic
- schema validation

PR #47, opened 2023-01-31
Don't validate JSON records against the table schema

genzgd:
JSON records don't normally have to match the table schema.  Unknown fields can be ignored (with a setting), and missing fields will use defaults.


genzgd:
I don't understand why this think it has a conflict?

mshustov (approved):
@genzgd could you create an issue to add tests later?

genzgd:
I think we generally need integration [tests](https://github.com/ClickHouse/clickhouse-kafka-connect/issues/4) for this section.  Unit tests won't mean much.

mshustov:
@genzgd agreed

PR #53, opened 2023-02-07
Git actions run tests

mzitnik:
## Summary
- Adding git actions


.github/workflows/build.yaml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'adopt'
+          architecture: x64
+      - name: Setup and execute Gradle 'test' task
```
mshustov, line 16: Let's rename the file to `test.yml` if we don't build

.github/workflows/build.yaml
```diff
@@ -0,0 +1,19 @@
+name: ClickHouse Kafka Connect Tests CI
+
+on: [push]
```
mshustov, line 3: Shouldn't we run tests on pull requests against `main`?

mzitnik: I want to run integration test when we have PR against main

mzitnik: I want to run integration test against main when we are doing PR, but i need first verify that github actions can handle it (memory & CPU consumption since it bootstrap a full env)

mshustov (approved):

PR #114, opened 2023-06-05
Updating the release process to split off the version (and support auto increment)

Paultagoras:
Testing on a different repo - this change allows us to increment the patch version manually (and also the minor version in a separate flow).


.github/workflows/increment-minor.yml
```diff
@@ -0,0 +1,34 @@
+name: Increment Minor
+
+on:
+  workflow_dispatch:
```
mzitnik, line 4: Can you add https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows as a link of a quick explanation of the events

.github/workflows/increment-patch.yml
```diff
@@ -0,0 +1,40 @@
+name: Increment Patch
+
+on:
```
mzitnik, line 3: Same here. Can you add https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows as a link of a quick explanation of the events

mzitnik (commented):
Can you explaine the flow here

mzitnik (approved):

PR #122, opened 2023-06-13
Adding code to switch to JSON input when default values are present

Paultagoras:
## Summary
This code checks to see if defaults are present in the DESCRIBE, and if so we'll end up switching to JSON handling instead.

### DIFF
--- a/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
+++ b/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
@@ -678,6 +678,33 @@ public class ClickHouseSinkTaskWithSchemaTest {
         chst.stop();
         assertEquals(sr.size(), countRows(chc, topic));
     }
+
+    @Test
+    public void testSchemaWithDefaults() {
+        Map<String, String> props = new HashMap<>();
+        props.put(ClickHouseSinkConnector.HOSTNAME, db.getHost());
+        props.put(ClickHouseSinkConnector.PORT, db.getFirstMappedPort().toString());
+        props.put(ClickHouseSinkConnector.DATABASE, "default");
+        props.put(ClickHouseSinkConnector.USERNAME, db.getUsername());
+        props.put(ClickHouseSinkConnector.PASSWORD, db.getPassword());
+        props.put(ClickHouseSinkConnector.SSL_ENABLED, "false");
+
+        ClickHouseHelperClient chc = createClient(props);
+
+        String topic = "default-value-table-test";
+        dropTable(chc, topic);
+        createTable(chc, topic, "CREATE TABLE `%s` ( `off16` Int16, default_value_data DateTime DEFAULT now() ) Engine = MergeTree ORDER BY off16");
+        // https://github.com/apache/kafka/blob/trunk/connect/api/src/test/java/org/apache/kafka/connect/data/StructTest.java#L95-L98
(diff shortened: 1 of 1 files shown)


src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -451,9 +464,19 @@ public void doInsertJson(List<Record> records) throws IOException, ExecutionExce
                 // write bytes into the piped stream
                 for (Record record: records ) {
                     if (record.getSinkRecord().value() != null ) {
-                        Map<String, Object> data = (Map<String, Object>) record.getSinkRecord().value();
-                        java.lang.reflect.Type gsonType = new TypeToken<HashMap>() {
-                        }.getType();
+                        Map<String, Object> data;
+                        if (record.getSinkRecord().value() instanceof Map) {
```
mzitnik, line 468: we can determine that with getSchemaType() instead of instanceof (less preferment)

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -451,9 +464,19 @@ public void doInsertJson(List<Record> records) throws IOException, ExecutionExce
...
+                            data = (Map<String, Object>) record.getSinkRecord().value();
+                        } else if (record.getSinkRecord().value() instanceof Struct) {
+                            data = new HashMap<>();
```
mzitnik, line 471: Can be pre-allocated with size in the constructor (do reallocation if needed default size 16)

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -451,9 +464,19 @@ public void doInsertJson(List<Record> records) throws IOException, ExecutionExce
...
+                            Struct struct = (Struct) record.getSinkRecord().value();
+                            for (Field field : struct.schema().fields()) {
+                                data.put(field.name(), struct.get(field));//Doesn't handle multi-level object depth
+                            }
+                        } else {
+                            throw new RuntimeException("Unsupported record type: " + record.getSinkRecord().value().getClass().getName());
+                        }
+                        java.lang.reflect.Type gsonType = new TypeToken<HashMap>() {}.getType();
```
mzitnik, line 479: Maybe we can use data.clear() to be less aggressive on the GC so we do not need to create data = new HashMap<>();
Member

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -132,7 +135,12 @@ public void doInsert(List<Record> records) {
             Record first = records.get(0);
             switch (first.getSchemaType()) {
                 case SCHEMA:
-                    doInsertRawBinary(records);
+                    try {
```
mzitnik, line 138: why handle it with exceptions if we have table.hasDefaults()

mzitnik (changes requested):

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
mzitnik: I think switch with 2 cases is cleaner than instanceof

mzitnik:
@Paultagoras, we are missing tests to covert the issue

mzitnik (approved):

PR #141, opened 2023-06-28
Update Utils.java

Paultagoras:
## Checklist

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/util/Utils.java
+++ b/src/main/java/com/clickhouse/kafka/connect/util/Utils.java
@@ -67,6 +67,7 @@ public class Utils {
                 case 252: // TOO_MANY_PARTS
                 case 285: // TOO_FEW_LIVE_REPLICAS
                 case 425: // SYSTEM_ERROR
+                case 1002: // UNKNOWN_EXCEPTION
                     throw new RetriableException(e);
                 default:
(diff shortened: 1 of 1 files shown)


src/main/java/com/clickhouse/kafka/connect/util/Utils.java
```diff
@@ -67,6 +67,7 @@ public static void handleException(Exception e) {
                 case 252: // TOO_MANY_PARTS
                 case 285: // TOO_FEW_LIVE_REPLICAS
                 case 425: // SYSTEM_ERROR
+                case 1002: // UNKNOWN_EXCEPTION
```
mzitnik, line 70: It will be interesting to understand in what cases in ClickHouse it can happen, but since retry is not forever we can add this status code

mzitnik (approved):

src/main/java/com/clickhouse/kafka/connect/util/Utils.java
mzitnik: @Paultagoras, I believe it happened to you in some cases. Can you expand?

mshustov: Is it safe to assume the error is retriable? I don't think so. In what circumstances did you get it?

PR #144, opened 2023-07-07
Feat/test multi shards

Bowenislandsong:
This PR enables connector to split evenly between shards of a ClickHouse database/keyspace. Test bed is able to create multiple shards and replicas.


Bowenislandsong:
Didn't mean to create against upstream main....

PR #149, opened 2023-07-13
Cleaning up and adding additional logs

Paultagoras:
## Summary
This adds additional logs to help us debug issues going forward, including trace details (like the raw data being sent) and what errors are being encountered.

It also tries to clean up some code that was flagged by the IDE for cleanliness.

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
+++ b/src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
@@ -484,11 +484,11 @@ public class ClickHouseWriter implements DBWriter{
                         
                         java.lang.reflect.Type gsonType = new TypeToken<HashMap>() {}.getType();
                         String gsonString = gson.toJson(data, gsonType);
-                        LOGGER.debug(String.format("topic [%s] partition [%d] offset [%d]",
+                        LOGGER.trace("topic {} partition {} offset {} payload {}",
                                 record.getTopic(),
                                 record.getRecordOffsetContainer().getPartition(),
-                                record.getRecordOffsetContainer().getOffset()));
-                        LOGGER.trace("payload {}", gsonString);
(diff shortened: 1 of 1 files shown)


src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
```diff
@@ -178,6 +190,16 @@ private static ConfigDef createConfigDef() {
                 ++orderInGroup,
                 ConfigDef.Width.MEDIUM,
                 "enable exactly once semantics.");
+        configDef.define(INSERT_QUORUM,
```
mzitnik, line 193: The range between 0 to 10 is problematic for big clusters. i think we can set it to auto and this will do the trick

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
```diff
@@ -45,10 +45,10 @@ private ClickHouseNode create() {
 
         String url = String.format("%s://%s:%d/%s", protocol, hostname, port, database);
 
-        LOGGER.info("url: " + url);
+        LOGGER.info("ClickHouse URL: " + url);
 
         if (username != null && password != null) {
-            LOGGER.info(String.format("Adding username [%s] password [%s]  ", username, Mask.passwordMask(password)));
+            LOGGER.debug(String.format("Adding username [%s]", username));
```
mzitnik, line 51: Why you removed the mask of the password from logging since in some cases it easy to troubleshot (i do not think we currently have a good error if the credentials are wrong)

mzitnik (changes requested):

mzitnik:
@Paultagoras Let's discuss the auto for quorum

mzitnik (approved):

PR #169, opened 2023-08-29
Quick optimization and adding custom settings

Paultagoras:
## Summary
The full set of clickhouse settings is now supported, and we handle overriding them all the same way - we provide a default, and someone can provide a different value.

I removed the 2 flags to support this cleaner flow, since it was only just introduced.

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
+++ b/src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
@@ -5,6 +5,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.Map;
 
 public class ClickHouseSinkConfig {
@@ -20,8 +21,6 @@ public class ClickHouseSinkConfig {
     public static final String TIMEOUT_SECONDS = "timeoutSeconds";
     public static final String RETRY_COUNT = "retryCount";
     public static final String EXACTLY_ONCE = "exactlyOnce";
-    public static final String SEND_PROGRESS_IN_HTTP_HEADERS = "sendProgressInHttpHeaders";
-    public static final String INPUT_FORMAT_SKIP_UNKNOWN_FIELDS = "inputFormatSkipUnknownFields";
     public static final String SUPPRESS_TABLE_EXISTENCE_EXCEPTION = "suppressTableExistenceException";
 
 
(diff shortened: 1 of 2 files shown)


src/exactlyOnceTest/java/com/clickhouse/ExactlyOnceTest.java
```diff
@@ -150,32 +150,35 @@ public static void tearDown() throws IOException, URISyntaxException, Interrupte
...
         Integer count = sendDataToTopic(topicName);
+        LOGGER.info("Expected Total: {}", count);
+        waitForCount(topicName);
 
-        Thread.sleep(60 * 1000);
-        LOGGER.info("Actual Total: {}", count);
-        int[] counts = clickhouseAPI.count(topicName);
-        if (counts != null) {
-            assertEquals(count, counts[1]);
+        int[] databaseCounts = clickhouseAPI.count(topicName);
```
mzitnik, line 162: Sorry for being so picky, but why return an array of ints and not Optional int  
This way, you can check if it has value and then extract it https://www.baeldung.com/java-optional

mzitnik, line 162: Sorry for being so picky, but why return an array of ints and not Optional int  
This way, you can check if it has value and then extract it https://www.baeldung.com/java-optional

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
```diff
@@ -84,11 +83,38 @@ public ClickHouseSinkConfig(Map<String, String> props) {
...
         retry = Integer.parseInt(props.getOrDefault(RETRY_COUNT, retryCountDefault.toString()));
         exactlyOnce = Boolean.parseBoolean(props.getOrDefault(EXACTLY_ONCE,"false"));
-        sendProgressInHttpHeaders = Boolean.parseBoolean(props.getOrDefault(SEND_PROGRESS_IN_HTTP_HEADERS,"true"));
-        inputFormatSkipUnknownFields = Boolean.parseBoolean(props.getOrDefault(INPUT_FORMAT_SKIP_UNKNOWN_FIELDS,"true"));
         suppressTableExistenceException = Boolean.parseBoolean(props.getOrDefault("suppressTableExistenceException","false"));
+
+        Map<String, String> clickhouseSettings = new HashMap<>();
+        String clickhouseSettingsString = props.getOrDefault("clickhouseSettings", "").trim();
+
+        if (!clickhouseSettingsString.isBlank()) {
```
mzitnik, line 91: Do we have the option to check if it is valid settings since if it is not valid we can log it to the user maybe he made a mistake in the setting (it is just an improvement to help us troubleshoot)

mzitnik (approved):

Issue #218, opened 2023-10-23
java.nio.ByteBuffer GSON Type


Paultagoras:
[2023-10-19 22:07:41,688] WARN [clickhouse-sink|task-0] Deciding how to handle exception: Failed making field 'java.nio.ByteBuffer#hb' accessible; either increase its visibility or write a custom TypeAdapter for its declaring type. (com.clickhouse.kafka.connect.util.Utils:51)
[2023-10-19 22:07:41,688] ERROR [clickhouse-sink|task-0] WorkerSinkTask{id=clickhouse-sink-0} Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted. Error: com.google.gson.JsonIOException: Failed making field 'java.nio.ByteBuffer#hb' accessible; either increase its visibility or write a custom TypeAdapter for its declaring type. (org.apache.kafka.connect.runtime.WorkerSinkTask:616)
java.lang.RuntimeException: com.google.gson.JsonIOException: Failed making field 'java.nio.ByteBuffer#hb' accessible; either increase its visibility or write a custom TypeAdapter for its declaring type.
at com.clickhouse.kafka.connect.util.Utils.handleException(Utils.java:95)
at com.clickhouse.kafka.connect.sink.db.ClickHouseWriter.doInsert(ClickHouseWriter.java:186)
at com.clickhouse.kafka.connect.sink.processing.Processing.doInsert(Processing.java:46)
at com.clickhouse.kafka.connect.sink.processing.Processing.doLogic(Processing.java:101)
at com.clickhouse.kafka.connect.sink.ProxySinkTask.put(ProxySinkTask.java:91)
at com.clickhouse.kafka.connect.sink.ClickHouseSinkTask.put(ClickHouseSinkTask.java:60)
at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:587)
at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:336)
at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:237)
at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:206)
at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:204)
at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:259)
[...]

PR #221, opened 2023-10-26
Adding ByteBuffer type adapter

Paultagoras:
This is meant to address https://github.com/ClickHouse/clickhouse-kafka-connect/issues/218


mzitnik:
@Paultagoras Did we manage to reproduce it in a test?

mzitnik (commented):
LGTM, but we need a test that produce this error to verify that we fixed it

PR #238, opened 2023-11-08
Update ClickHouseHelperClient.java

Paultagoras:
This fixes the "https" error we were seeing - our unit tests (and integration tests, technically) don't use SSL and didn't run into this.

I'll look at augmenting them.


mzitnik:
@Paultagoras, First, I do not understand the fix. Could we please sync on that? Second, our integration tests work with ClickHouse Cloud so it uses SSL. 
Maybe it sign that we need also to cover HTTPS on our unit tests.

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
```diff
@@ -119,9 +119,11 @@ public ClickHouseResponse query(String query, ClickHouseFormat clickHouseFormat)
         int retryCount = 0;
         ClickHouseException ce = null;
         while (retryCount < retry) {
-            try (ClickHouseClient client = ClickHouseClient.newInstance(ClickHouseProtocol.HTTP);
+            try (ClickHouseClient client = ClickHouseClient.builder()
```
mzitnik, line 122: What is the reason that we changed the way we work.

mzitnik (approved):

PR #253, opened 2023-11-19
Limiting the amount of retries

Paultagoras:
This tweaks the retry mechanism to enforce a retry limit - out of the box, Kafka Connect will retry any RetriableException it receives (no exceptions, pun intended).

Note: We have to use the existing RetriableException - I tested using a custom retry mechanism and you run into a timeout issue with Kafka and the consumer groups (You need to poll periodically to keep alive the consumer). This way we still limit the attempts, and it's still mostly Out-Of-The-Box.

I also tweaked some of the exceptions so that they bubble up to the single point in put() rather than at two places - it should help reduce some logging confusion.


mzitnik:
QQ: Do you want to have an option to enable unlimited retries? 
Do you think putAttempts should be [volatile](https://www.baeldung.com/java-volatile) let's think if it is the case here.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -69,7 +69,7 @@ public boolean start(ClickHouseSinkConfig csc) {
                 .setPassword(csc.getPassword())
                 .sslEnable(csc.isSslEnabled())
                 .setTimeout(csc.getTimeout())
-                .setRetry(csc.getRetry())
+                .setRetry(csc.getMaxRetry())
```
mzitnik, line 72: you have changed all variables to maxRetry should we also change it here

PR #254, opened 2023-11-20
Updated retry mechanics and additional logging

Paultagoras:
This is based on discussions - the retry logic allows for infinite retries by default but changing the property caps the number of attempts.

If max retries is hit, an exception will be thrown (regardless of errorTolerance) - this is to avoid cycling through the topic with temporary errors and moving beyond them (as opposed to... say data exceptions, for instance)

This also adds some more explicit logging around exceptions to try and capture more useful data when debugging errors.

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
+++ b/src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
@@ -39,7 +39,7 @@ public class ClickHouseSinkConfig {
     public static final String PROXY_PORT = "proxyPort";
     public static final String ZK_PATH = "zkPath";
     public static final String ZK_DATABASE = "zkDatabase";
-    public static final String MAX_RETRIES = "maxRetries";
+    public static final String MAX_RETRIES = "max.retries";
 
 
     


src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
```diff
@@ -39,6 +39,7 @@ public class ClickHouseSinkConfig {
     public static final String PROXY_PORT = "proxyPort";
     public static final String ZK_PATH = "zkPath";
     public static final String ZK_DATABASE = "zkDatabase";
+    public static final String MAX_RETRIES = "maxRetries";
```
mshustov, line 42: Why does the Sink use `retryCount` config instead of the standard `max.retries`? 
see [elasticsearch connector](https://docs.confluent.io/kafka-connectors/elasticsearch/current/overview.html#automatic-retries), [JDBC connector](https://docs.confluent.io/kafka-connectors/jdbc/current/sink-connector/sink_config_options.html#retries)

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
```diff
@@ -219,6 +222,7 @@ public ClickHouseSinkConfig(Map<String, String> props) {
         this.proxyPort = Integer.parseInt(props.getOrDefault(PROXY_PORT, "-1"));
         this.zkPath = props.getOrDefault(ZK_PATH, "/kafka-connect");
         this.zkDatabase = props.getOrDefault(ZK_DATABASE, "connect_state");
+        this.maxRetries = Integer.parseInt(props.getOrDefault(MAX_RETRIES, "-1"));
```
mshustov, line 225: Do we need an additional `retry.backoff.ms` parameter?

PR #256, opened 2023-11-22
Logging updates

Paultagoras:
These are just the logging updates, separate from the retry mechanics.


src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -65,7 +65,7 @@ public void put(Collection<SinkRecord> records) {
         } catch (Exception e) {
             LOGGER.trace("Passing the exception to the exception handler.");
             boolean errorTolerance = clickHouseSinkConfig != null && clickHouseSinkConfig.getErrorsTolerance();
-            Utils.handleException(e, errorTolerance);
+            Utils.handleException(e, errorTolerance, records);
```
mzitnik, line 68: If we could somehow populate the offsets and not iterate all records

mzitnik: That I have seen in the handleException implementation

mzitnik: wdyt?

mzitnik (approved):
Approving not block

Issue #278, opened 2023-12-07
NullPointerException if in array type go null value with avro


oleg-savko:
Use 1.0.7 version

Connector throw NullPointerException:
```
{
name: "...",
connector: {
state: "RUNNING",
worker_id: "connect:8083"
},
tasks: [
{
id: 0,
state: "FAILED",
worker_id: "connect:8083",
trace: "org.apache.kafka.connect.errors.ConnectException: Exiting WorkerSinkTask due to unrecoverable exception.
	at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:628)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:340)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:238)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:207)
	at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:229)
	at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:284)
	at org.apache.kafka.connect.runtime.isolation.Plugins.lambda$withClassLoader$1(Plugins.java:181)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.RuntimeException: java.lang.NullPointerException
	at com.clickhouse.kafka.connect.util.Utils.handleException(Utils.java:124)
	at com.clickhouse.kafka.connect.sink.ClickHouseSinkTask.put(ClickHouseSinkTask.java:68)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:593)
	... 11 more
Caused by: java.lang.NullPointerException
	at com.clickhouse.kafka.connect.sink.data.StructToJsonMap.toJsonMap(StructToJsonMap.java:67)
[...]

PR #279, opened 2023-12-08
Fixed exception for when an array type column gets null value

mlivirov:
## Summary
Fixed exception for when an array type column gets null value

Issue: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/278

## Checklist

### DIFF
--- a/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
+++ b/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
@@ -120,6 +120,22 @@ public class ClickHouseSinkTaskWithSchemaTest {
     }
 
     @Test
+    public void nullableArrayTypeTest() {
+        Map<String, String> props = getTestProperties();
+        ClickHouseHelperClient chc = createClient(props);
+
+        String topic = "nullable_array_string_table_test";
+        ClickHouseTestHelpers.dropTable(chc, topic);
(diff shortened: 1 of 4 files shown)


src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
```diff
@@ -119,6 +119,22 @@ public void mapTypesTest() {
...
+        ClickHouseHelperClient chc = createClient(props);
+
+        String topic = "nullable_array_string_table_test";
+        ClickHouseTestHelpers.dropTable(chc, topic);
+        ClickHouseTestHelpers.createTable(chc, topic, "CREATE TABLE %s ( `off16` Int16, `arr` Array(String)  ) Engine = MergeTree ORDER BY off16");
+        Collection<SinkRecord> sr = SchemaTestData.createNullableArrayType(topic, 1);
+
+        ClickHouseSinkTask chst = new ClickHouseSinkTask();
+        chst.start(props);
+        assertThrowsExactly(RuntimeException.class, () -> chst.put(sr), "An attempt to write null into not nullable column 'arr'");
```
oleg-savko, line 134: It should throw exception when null value come to array type, because in this way it cant be resolved.
It should pass empty array! 
In ClickHouse null value equal empty array for Array type.
```
with '{"x": null}' as json
select JSONExtractArrayRaw(json, 'x');
-----
result []
```
image

mlivirov: Yeah, this is a small PR that just fixes current behaviour. So instead of Null Pointer Exception something more meaningful will appear in the log. 

I guess we gonna need to wait till maintainers decided what to do with the bug you've submitted, cuz I don't feel comfortable making such changes.

PR #288, opened 2023-12-19
Send an empty array when field is null

Paultagoras:
So this handles the null schema array issue, with a bit of selective logic.

If the subtype is nullable (e.g. Array(Nullable(String))) then we will pass an empty array if the array is sent null.
If the subtype is NOT nullable (e.g. Array(String)) then we through an exception as normal for "null when not nullable".

Since Array itself cannot be nullable, I thought this was a good compromise and consistent with the logic - does it make sense?

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
+++ b/src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
@@ -350,7 +350,7 @@ public class ClickHouseWriter implements DBWriter {
                 BinaryStreamUtils.writeNonNull(stream);
             }
             if (!col.isNullable() && value.getObject() == null) {
-                if (colType == Type.ARRAY && col.getSubType().isNullable())
+                if (colType == Type.ARRAY)
                     BinaryStreamUtils.writeNonNull(stream);
                 else
                     throw new RuntimeException(String.format("An attempt to write null into not nullable column '%s'", col.getName()));
--- a/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
+++ b/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
@@ -131,8 +131,10 @@ public class ClickHouseSinkTaskWithSchemaTest {
 
         ClickHouseSinkTask chst = new ClickHouseSinkTask();
         chst.start(props);
-        assertThrowsExactly(RuntimeException.class, () -> chst.put(sr), "An attempt to write null into not nullable column 'arr'");
+        chst.put(sr);
         chst.stop();
+
+        assertEquals(sr.size(), ClickHouseTestHelpers.countRows(chc, topic));
     }
 
     @Test


src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
```diff
@@ -135,6 +135,24 @@ public void nullableArrayTypeTest() {
         chst.stop();
     }
 
+    @Test
+    public void nullableArrayTypeTest() {
```
mshustov, line 139: Should we add the same logic for Tuple?

oleg-savko:
> So this handles the null schema array issue, with a bit of selective logic.
> 
> If the subtype is nullable (e.g. Array(Nullable(String))) then we will pass an empty array if the array is sent null. If the subtype is NOT nullable (e.g. Array(String)) then we through an exception as normal for "null when not nullable".
> 
> Since Array itself cannot be nullable, I thought this was a good compromise and consistent with the logic - does it make sense?

I think it should be empty array in both case.

For-example I stuck with this, when try create replication from postgres to clickhouse, via `debezium + clickhouse-kafka-connect` , current with `avro` and `null arrays` in postgres, you have a unresolvable problem,
and should return to `json` from `avro`:


In ClickHouse with json its work fine, both empty:
```
with '{"x": null}' as json
select JSONExtract(json, 'x', 'Array(String)'),
       JSONExtract(json, 'x', 'Array(Nullable(String))');
```

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -403,18 +401,23 @@ private void doWriteCol(Record record, Column col, ClickHousePipedOutputStream s
...
                     List<?> arrObject = (List<?>) value.getObject();
-                    int sizeArrObject = arrObject.size();
-                    BinaryStreamUtils.writeVarInt(stream, sizeArrObject);
-                    arrObject.forEach(v -> {
-                        try {
-                            if (col.getSubType().isNullable() && v != null) {
-                                BinaryStreamUtils.writeNonNull(stream);
+
+                    if (arrObject == null) {
+                        doWritePrimitive(colType, value.getFieldType(), stream, new ArrayList<>());
```
mzitnik, line 406: Do we cover this by test?

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -403,18 +401,23 @@ private void doWriteCol(Record record, Column col, ClickHousePipedOutputStream s
...
+                        int sizeArrObject = arrObject.size();
+                        BinaryStreamUtils.writeVarInt(stream, sizeArrObject);
+                        arrObject.forEach(v -> {
+                            try {
+                                if (col.getSubType().isNullable() && v != null) {
+                                    BinaryStreamUtils.writeNonNull(stream);
+                                }
+                                doWritePrimitive(col.getSubType().getType(), value.getFieldType(), stream, v);
+                            } catch (IOException e) {
+                                throw new RuntimeException(e);
```
mzitnik, line 417: Do not know what type of exceptions we can have here, but should we cases that we want to retry or put it in DLQ

mzitnik (approved):

PR #291, opened 2023-12-28
fix 'overlapping' logic

ne1r0n:
## Summary

1. Excluded excess data processing when state = `BEFORE_PROCESSING'
2. Fixed order of chunks in overlapping range cases
3. Prevented insertion of the 1 chunk in overlapping range case when state = `BEFORE_PROCESSING' and 1 chunk boundaries are not the same as state
4. Fixed and extended different range overlapping cases taking into account that records are trimmed before processing, for example:
```
State                 Actual                 Current                 After fix
[2, 10]               [0, 11]                ZERO                    OVER_LAPPING
[2, 10]               [0, 10]                ZERO                    SAME
[2, 10]               [1, 11]                ERROR                   OVER_LAPPING
[2, 10]               [1, 10]                ERROR                   SAME
```

## Checklist

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/sink/kafka/RangeContainer.java
+++ b/src/main/java/com/clickhouse/kafka/connect/sink/kafka/RangeContainer.java
@@ -51,23 +51,23 @@ public class RangeContainer extends TopicPartitionContainer {
         long actualMinOffset = rangeContainer.getMinOffset();
         long actualMaxOffset = rangeContainer.getMaxOffset();
 
-        // SAME State [0,10] Actual [0,10]
-        if (maxOffset == actualMaxOffset && minOffset == actualMinOffset)
+        // SAME State [0, 10] Actual [0, 10]
+        if (actualMaxOffset == maxOffset && actualMinOffset <= minOffset)
             return RangeState.SAME;
-        // NEW State [0,10] Actual [11,20]
+        // NEW State [0, 10] Actual [11, 20]
         if (actualMinOffset > maxOffset)
             return RangeState.NEW;
-        // CONTAINS [0,10] Actual [1, 10]
+        // CONTAINS [0, 10] Actual [1, 10]
         if (actualMaxOffset <= maxOffset && actualMinOffset >= minOffset)
             return RangeState.CONTAINS;
-        // ZEROED [10, 20] Actual [0, 10]
-        if (actualMinOffset < minOffset && actualMinOffset == 0)
-            return RangeState.ZERO;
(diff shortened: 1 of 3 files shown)


mzitnik:
@ne1r0n can you provide a test with content that cover this?

ne1r0n:
> @ne1r0n can you provide a test with content that cover this?

@mzitnik it would require a lot of effort from my side as I'm not a professional java programmer.

All I am trying to fix is the causes of missed messages when using a connector in production-like environment.

For example, in rare cases there could be issues with committing offsets, something like:
`org.apache.kafka.clients.consumer.CommitFailedException: Offset commit cannot be completed since the consumer is not part of an active group for auto partition assignment; it is likely that the consumer was kicked out of the group.`

After that failure consumer could read more messages from kafka than previously so batch of data would contain part that already was inserted into clickhouse and a new part that wasn't. Currently connector will skip the whole batch.
According to the design document, it is either `overlapping` or the combination of `contains` and `new`, where the `new` should be inserted.


It make sense to implement an integration test that cover above-mentioned this scenario.


Another test could be for `splitRecordsByOffset` checking the order of parts it returns.

mzitnik:
@ne1r0n We will discuss it internally since I want a test that covers it.

ne1r0n:
I've added 2 tests this PR is trying to fix, please have a look.

ne1r0n:
@mzitnik , @Paultagoras I will be grateful for any updates or comments

ne1r0n:
Hi @Paultagoras,
currently the part shown in the picture will be lost
image

ne1r0n:
@Paultagoras yep, I saw it, thanks. It solved part of the problem.

Now let's imagine the situation like in `ProcessPartialOverlappingBeforeProcessingTest` test.
We have 2 batches with overlapping ranges and BEFORE_PROCESSING state after 1 insert.
     0 to  599 - 1 batch
 350 to 999 - 2 batch

Connector will do the following:
1. insert the 1 batch of 600 records
2. split the second batch into two parts: 350 to 599 and 600 to 999
3. insert the 1st part 350 to 599 in BEFORE_PROCESSING case
4. insert the 2nd part 600 to 999 in BEFORE_PROCESSING case
5. insert the 2nd part 600 to 999 in AFTER_PROCESSING case

There are 2 issues:
Step 3 correspond to **contains** case from the design document and should be processed accordingly.
Step 5 extra insert of the second part, here `deduplication_token` will fix that, but I think there should be `break` before `case AFTER_PROCESSING:`

PR #327, opened 2024-02-21
Adding String support and tweaking validation

Paultagoras:
Tweaking the FIXED_STRING code to support String as a source type, and removing the validation from the connector so that Clickhouse can do the validation instead (which it handles)

### DIFF
--- a/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
+++ b/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
@@ -11,6 +11,8 @@ import org.apache.kafka.connect.sink.SinkRecord;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
```diff
@@ -439,18 +441,18 @@ public void schemaWithFixedStringMismatchTest() {
         ClickHouseHelperClient chc = createClient(props);
 
         String topic = "fixed-string-mismatch-table-test";
-        int fixedStringSize = RandomUtils.nextInt(1, 100);
+        int fixedStringSize = RandomUtils.nextInt(2, 100);
```
mzitnik, line 444: Can we log it so if we encounter any issue here, we can reproduce easily

mzitnik (changes requested):

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
mzitnik: In the test itself, if for some reason it fails and we want to reproduce it, we need to have the exact `fixedStringSize`

mzitnik (approved):

PR #352, opened 2024-03-21
Adjusting logic to throw an exception rather than silently fail

Paultagoras:
So with the ERROR condition it would silently discard records - this way, it throws an exception (and the normal error process happens with the DLQ).

That's also true of the CONTAINS scenario - rather than send to the DLQ silently, we use the normal exception handling (so that if the DLQ ISN'T defined, it will trigger and error condition).

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/sink/processing/Processing.java
+++ b/src/main/java/com/clickhouse/kafka/connect/sink/processing/Processing.java
@@ -140,14 +140,9 @@ public class Processing {
                         stateProvider.setStateRecord(new StateRecord(topic, partition, rangeContainer.getMaxOffset(), rangeContainer.getMinOffset(), State.AFTER_PROCESSING));
                         break;
                     case CONTAINS: // The state contains the given records
-                        LOGGER.warn(String.format("Records seemingly missing compared to prior batch for topic [%s] partition [%s].", topic, partition));
                         // Do nothing - write to dead letter queue
-                        LOGGER.warn("Due to CONTAINS state, sending [{}] records to DLQ for topic: {}, partition: {}, minOffset: {}, maxOffset: {}",
-                                records.size(), topic, partition, rangeContainer.getMinOffset(), rangeContainer.getMaxOffset());
-                        records.forEach( r ->
-                                Utils.sendTODlq(errorReporter, r, new DuplicateException(String.format(record.getTopicAndPartition())))
-                        );
-                        break;
+                        throw new RuntimeException(String.format("State CONTAINS given [%s] records for topic: [%s], partition: [%s], minOffset: [%s], maxOffset: [%s]",
+                                records.size(), topic, partition, rangeContainer.getMinOffset(), rangeContainer.getMaxOffset()));
                     case OVER_LAPPING:
                         // spit it to 2 inserts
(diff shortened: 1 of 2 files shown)


src/main/java/com/clickhouse/kafka/connect/sink/processing/Processing.java
```diff
@@ -140,14 +140,9 @@ public void doLogic(List<Record> records) throws IOException, ExecutionException
...
                     case CONTAINS: // The state contains the given records
-                        LOGGER.warn(String.format("Records seemingly missing compared to prior batch for topic [%s] partition [%s].", topic, partition));
                         // Do nothing - write to dead letter queue
-                        LOGGER.warn("Due to CONTAINS state, sending [{}] records to DLQ for topic: {}, partition: {}, minOffset: {}, maxOffset: {}",
-                                records.size(), topic, partition, rangeContainer.getMinOffset(), rangeContainer.getMaxOffset());
-                        records.forEach( r ->
-                                Utils.sendTODlq(errorReporter, r, new DuplicateException(String.format(record.getTopicAndPartition())))
-                        );
-                        break;
+                        throw new RuntimeException(String.format("State CONTAINS given [%s] records for topic: [%s], partition: [%s], minOffset: [%s], maxOffset: [%s]",
```
mzitnik, line 144: Will this cause the connector to stop? And then, the user needs to make a manual adjustment to offsets to continue ?

src/main/java/com/clickhouse/kafka/connect/sink/processing/Processing.java
```diff
@@ -162,8 +157,8 @@ public void doLogic(List<Record> records) throws IOException, ExecutionException
                         stateProvider.setStateRecord(new StateRecord(topic, partition, rightRangeContainer.getMaxOffset(), rightRangeContainer.getMinOffset(), State.AFTER_PROCESSING));
                         break;
                     case ERROR:
-                        LOGGER.warn(String.format("State mismatch for topic [%s] partition [%s].", topic, partition));
-                        break;
+                        throw new RuntimeException(String.format("State MISMATCH given [%s] records for topic: [%s], partition: [%s], minOffset: [%s], maxOffset: [%s]",
```
mzitnik, line 160: Will this cause the connector to stop? And then, the user needs to make a manual adjustment to offsets to continue ?

src/main/java/com/clickhouse/kafka/connect/sink/processing/Processing.java
```diff
@@ -195,7 +190,8 @@ public void doLogic(List<Record> records) throws IOException, ExecutionException
                         stateProvider.setStateRecord(new StateRecord(topic, partition, rightRangeContainer.getMaxOffset(), rightRangeContainer.getMinOffset(), State.AFTER_PROCESSING));
                         break;
                     case ERROR:
-                        LOGGER.warn(String.format("State mismatch for topic [%s] partition [%s]", topic, partition));
+                        throw new RuntimeException(String.format("State MISMATCH given [%s] records for topic: [%s], partition: [%s], minOffset: [%s], maxOffset: [%s]",
```
mzitnik, line 193: Will this cause the connector to stop? And then, the user needs to make a manual adjustment to offsets to continue ?

mzitnik:
@Paultagoras, maybe we can make this configurable to give an option to user to decide if he want to stop connector or not

PR #354, opened 2024-03-26
Update Utils.java to remove 1002 from exceptions list

Paultagoras:
Since this is a catch-all, we shouldn't retry...


mzitnik:
@Paultagoras I think we can close the PR

mzitnik (approved):

PR #376, opened 2024-04-14
Change to async_insert=0

mzitnik:
## Summary
Change to async_insert=0

## Checklist


ne1r0n:
It seems that after this change it will be impossible to enable async insert via profile in server config or via SETTINGS PROFILE, but only via connect parameter `clickhouseSettings`. Is it intended? In some cases it is not very convenient.

mzitnik:
@ne1r0n OK, i will create a PR with `async_insert=0` as default but with an option to change it

PR #385, opened 2024-05-14
Pass modified database to doConvert if split db is true

abhishekgahlot2:
## Checklist


abhishekgahlot2:
I think we need tests for this, because looks like this case was missing from test.

PR #392, opened 2024-05-17
Adding some data validation

Paultagoras:
This isn't (yet) a super thorough validation but it's much more than we used to!

### DIFF
--- a/src/test/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
+++ b/src/test/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
@@ -1,22 +1,29 @@
 package com.clickhouse.kafka.connect.sink.helper;
 
 import com.clickhouse.client.*;
+import com.clickhouse.data.ClickHouseColumn;
 import com.clickhouse.data.ClickHouseFormat;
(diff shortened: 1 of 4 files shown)


src/test/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
```diff
@@ -142,6 +146,58 @@ public static int countRowsWithEmojis(ClickHouseHelperClient chc, String tableNa
...
+            List<String> results = new ArrayList<>();
+            LOGGER.info(response.records().toString());
+            response.records().forEach(r -> {
+                String gsonString = r.getValue(0).asString().replace("'","").replace(" ","").replace("\\u003d",":");
+                Map<String, String> resultMap = new TreeMap<>((Map<String, String>) gson.fromJson(gsonString, new TypeToken<Map<String, String>>() {}.getType()));
+                results.add(gson.toJson(resultMap));
+            });
+
+            for (String record : records) {
+                if (results.get(0).equals(record)) {
```
mzitnik, line 185: We are checking only 1 record, it is interesting to know we expect the retrieve same orders as we inserted what will happen when we are insetting using several workers i think it mainly can impact on our integration tests

mzitnik (approved):

PR #402, opened 2024-06-03
Update ClickHouseWriter.java

Paultagoras:
We were relying on the default encoding, which is obviously not ideal...


mzitnik (approved):

mzitnik:
@Paultagoras I think we can add another method to `BinaryStreamUtils.writeBytes` that does it internally also so other clients will benefit instead carrying this logic every where

mzitnik:
I also notice that we have some cloud tests that a failing lets fix them

mzitnik:
@Paultagoras let's write a test for it.

PR #404, opened 2024-06-05
Tweak to return null instead of throw exception

Paultagoras:
Currently "Nested" columns that are created using flatten_nested=0 cause an exception - at least for now I've set it to warn and then ignore that table, though we could explore if it's possible to handle that condition better.


mzitnik (approved):
Do you think we can write a tests for it?

mzitnik:
@Paultagoras Do you think we can write a tests for it?

PR #405, opened 2024-06-07
Fix invalid `ON CLUSTER` clause

junegunn:
## Summary

This commit fixes the following error,

    java.lang.RuntimeException: com.clickhouse.client.ClickHouseException:
    Code: 62. DB::Exception: Syntax error: failed at position 109 ('ON'):
    ON CLUSTER ******* ENGINE=KeeperMap('/kafka-connect') PRIMARY KEY
    `key`;. Expected one of: storage definition, ENGINE, PARTITION BY,
    PRIMARY KEY, ORDER BY, SAMPLE BY, TTL, EMPTY AS, AS, COMMENT, INTO
    OUTFILE, FORMAT, SETTINGS, end of query. (SYNTAX_ERROR) (version
    24.4.1.2088 (official build))

which is caused by an incorrect placement of `ON CLUSTER` clause. It should immediately follow the table name.

https://clickhouse.com/docs/en/sql-reference/statements/create/table


junegunn:
> Do you happen to have a sample of the CREATE statement

It's on the error message I posted above. `ON CLUSTER ******* ENGINE=KeeperMap`.

`ON CLUSTER` shouldn't be there.

abhishekgahlot2:
@Paultagoras in your sql 

```
CREATE TABLE IF NOT EXISTS all_hits ON CLUSTER default (p Date, i Int32) ENGINE = SharedMergeTree('/clickhouse/tables/{uuid}/{shard}', '{replica}') ORDER BY p
```

ON CLUSTER is before column definition but in current main branch its defined after the column definition which seems to be creating this error.

This works
```
CREATE TABLE IF NOT EXISTS `my_table1` ON CLUSTER CLUSTER_NAME
(`key1` String, minOffset1 BIGINT, maxOffset1 BIGINT, state String) 
 ENGINE=KeeperMap('/kafka-connect2') PRIMARY KEY `key1`;
```


This doesn't
```
CREATE TABLE IF NOT EXISTS `my_table`
(`key1` String, minOffset1 BIGINT, maxOffset1 BIGINT, state String) 
 ON CLUSTER CLUSTER_NAME
 ENGINE=KeeperMap('/kafka-connect1') PRIMARY KEY `key1`;
```

PR #410, opened 2024-06-28
Adjusting the table mapping code to account for additional databases

Paultagoras:
Because we allow folks to specify database/topic splits (drawing on additional databases) we have to update the mapping functions to support that. This way, the list will be updated if a table is missing (then caching it for future calls).


mzitnik (approved):
There is the situation that we do not know in advance what databases we will insert data into and we need to scan the list of tables for a specific database that was not selected in our configuration do we cover this situation?

PR #413, opened 2024-07-10
Adding additional String parsing for DateTime64

Paultagoras:
So Date is precise to milliseconds, where folks sometimes want nanoseconds. To help address this, I've adjusted the String parsing to allow for custom patterns and then transmitting that based on the DateTime64 precision.

### DIFF
--- a/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
+++ b/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
@@ -34,6 +34,7 @@ public class ClickHouseBase {
             database = String.format("kafka_connect_test_%s", System.currentTimeMillis());
         }
         if (isCloud) {
+            initialPing();
             return;
         }
         db = new ClickHouseContainer(ClickHouseTestHelpers.CLICKHOUSE_DOCKER_IMAGE);
@@ -119,6 +120,45 @@ public class ClickHouseBase {
             throw new RuntimeException(e);
         }
 
+    }
+
+    protected static void initialPing() {
+        ClickHouseSinkConfig csc = new ClickHouseSinkConfig(new ClickHouseBase().createProps());
+
+        String hostname = csc.getHostname();
+        int port = csc.getPort();
+        String username = csc.getUsername();
+        String password = csc.getPassword();
+        boolean sslEnabled = csc.isSslEnabled();
+        int timeout = csc.getTimeout();
+
+        ClickHouseHelperClient tmpChc = new ClickHouseHelperClient.ClickHouseClientBuilder(hostname, port, csc.getProxyType(), csc.getProxyHost(), csc.getProxyPort())
+                .setDatabase("default")
+                .setUsername(username)
+                .setPassword(password)
+                .sslEnable(sslEnabled)
+                .setTimeout(timeout)
+                .setRetry(csc.getRetry())
+                .build();
+
+        boolean ping;
+        int retry = 0;
+        do {
+            LOGGER.info("Pinging ClickHouse server to warm up for tests...");
+            ping = tmpChc.ping();
+            if (!ping) {
+                LOGGER.info("Unable to ping ClickHouse server, retrying... {}", retry);
+                try {
+                    Thread.sleep(5000);
+                } catch (InterruptedException e) {
+                    e.printStackTrace();
+                }
+            }
+        } while (!ping && retry++ < 10);
+
+        if (!ping) {
+            throw new RuntimeException("Unable to ping ClickHouse server...");
+        }
     }
 
     protected static void dropDatabase(String database) {


src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -302,18 +306,31 @@ private void doWriteDates(Type type, ClickHousePipedOutputStream stream, Data va
                 if (value.getFieldType().equals(Schema.Type.INT64)) {
                     if (value.getObject().getClass().getName().endsWith(".Date")) {
                         Date date = (Date) value.getObject();
-                        long time = date.getTime();
-                        BinaryStreamUtils.writeInt64(stream, time);
```
mzitnik, line 306: Reducing code we love it

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
```diff
@@ -521,6 +524,15 @@ private static ConfigDef createConfigDef() {
                 ++orderInGroup,
                 ConfigDef.Width.SHORT,
                 "Bypass RowBinary format.");
+        configDef.define(DATE_TIME_FORMAT,
```
mzitnik, line 527: Can we provide an example for it

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -302,18 +306,31 @@ private void doWriteDates(Type type, ClickHousePipedOutputStream stream, Data va
...
-                        long milliSeconds = zonedDateTime.toInstant().toEpochMilli();
-                        long microSeconds = TimeUnit.MICROSECONDS.convert(seconds, TimeUnit.SECONDS) + zonedDateTime.get(ChronoField.MICRO_OF_SECOND);
-                        long nanoSeconds = TimeUnit.NANOSECONDS.convert(seconds, TimeUnit.SECONDS) + zonedDateTime.getNano();
+                        long seconds;
+                        long milliSeconds;
+                        long microSeconds;
+                        long nanoSeconds;
+
+                        if (!csc.getDateTimeFormat().isEmpty()) {
+                            DateTimeFormatter formatter = DateTimeFormatter.ofPattern(csc.getDateTimeFormat());
```
mzitnik, line 321: Maybe we can initialize the `formatter` only once -> also we need to explain that we only support one format (how we plan to deal if we have different formats for different variables )

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
```diff
@@ -266,6 +267,39 @@ public void supportZonedDatesStringTest() {
     }
 
 
+    @Test
+    public void supportFormattedDatesStringTest() {
+        Map<String, String> props = createProps();
```
mzitnik, line 272: I may be missing something here. I think it needs to be a data validation test (that our conversion is correct)?

src/test/java/com/clickhouse/kafka/connect/sink/helper/SchemaTestData.java
```diff
@@ -907,6 +908,40 @@ public static Collection<SinkRecord> createZonedTimestampConversions(String topi
...
+                .field("off16", Schema.INT16_SCHEMA)
+                .field("format_date", Schema.STRING_SCHEMA)
+                .build();
+
+
+        List<SinkRecord> array = new ArrayList<>();
+        LongStream.range(0, totalRecords).forEachOrdered(n -> {
+            Struct value_struct = new Struct(NESTED_SCHEMA)
+                    .put("off16", (short)n)
+                    .put("format_date", LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSSSSS")));
```
mzitnik, line 926: Maybe we should have some more fields (year, day, ...) so we will check our conversion

mzitnik (changes requested):

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
```diff
@@ -245,6 +248,17 @@ public ClickHouseSinkConfig(Map<String, String> props) {
         this.dbTopicSplitChar = props.getOrDefault(DB_TOPIC_SPLIT_CHAR, "");
         this.keeperOnCluster = props.getOrDefault(KEEPER_ON_CLUSTER, "");
         this.bypassRowBinary = Boolean.parseBoolean(props.getOrDefault("bypassRowBinary", "false"));
+        this.dateTimeFormats = new HashMap<>();
+        String dateTimeFormatsString = props.getOrDefault(DATE_TIME_FORMAT, "").trim();
+        if (!dateTimeFormatsString.isBlank()) {
+            String [] stringSplit = dateTimeFormatsString.split(";");
```
mzitnik, line 254: Using `;` as a separator might be fine, but I was thinking more as `"xxx={format1}","yyy={format2}","zzz={format3}"`
Split `","` truncate `"` and to parse it inside i think it is more error-prone

mzitnik (approved):
I'am approving but refer to my comment (maybe we also need to aline all other place that we are doing those type of parsing just to be consistent)

Issue #422, opened 2024-07-30
ClassCastException due to invalid field value to column name mapping for Tuples


arnay-allen:
### Describe the bug
The **doWriteColValue** method in **ClickHouseWriter.java** in case of tuples uses
``` 
Streams.zip(
    col.getTupleFields().stream(), value.getFields().stream(), Tuples:: of
).forEach((fields) - > {
    Column column = fields.getT1();
    Field field = fields.getT2();

    Data innerData = (Data) jsonMapValues.get(field.name());
    try {
        doWriteColValue(column, stream, innerData, defaultsSupport);
    } catch (IOException e) {
        throw new RuntimeException(e);
    }
});
````
to get the field value from `jsonMapValues` for a given column. However, since the `col.getTupleFields()` and `value.getFields()` are not sorted by the column names, a mismatch occurs when fetching column and field from the Streams.zip, which ultimately leads to ClassCastException in `doWritePrimitive`
### Steps to reproduce
1. Create a clickhouse table which uses Tuples having different datatypes like Int32, Int64, Float64, String, etc.
2. Create a [Kafka Connect MongoDB source connector](https://www.mongodb.com/docs/kafka-connector/current/) to push data from collection to Kafka Topic.
3. Create a [Kafka Connect Clickhouse Sink Connector](https://clickhouse.com/docs/en/integrations/kafka/clickhouse-kafka-connect-sink) to read the messages on the Kafka topic and insert into Clickhouse table

### Error log
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String

PR #423, opened 2024-07-30
Issue #422 | Replacing Streams.zip with forEach on TupleFields

arnay-allen:
#422
Since the two lists in Streams.zip were not sorted wrt to column names, a mismatch occurred between names & values causing ClassCastException in `doWritePrimitive`


arnay-allen:
fixes #422

arnay-allen:
@Paultagoras added a unit test catering to tuples

PR #430, opened 2024-08-18
Add some initial code to take the key and store it in the value

Paultagoras:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 1.1.4
 * Bugfix to address field value to column name mapping for Tuples
+* Adding a KeyToValue transformation to allow for key to be stored in a separate column in ClickHouse
 
 # 1.1.3
 * Update to java-client 0.6.3


src/main/java/com/clickhouse/kafka/connect/transforms/KeyToValue.java
```diff
@@ -0,0 +1,81 @@
...
+        LOGGER.debug("Old Key: {}, Old Value: {}", record.key(), record.value());
+        if (record.valueSchema() == null) {
+            return applySchemaless(record);
+        } else {
+            return applyWithSchema(record);
+        }
+    }
+
+    private R applySchemaless(R record) {
+        final Map<String, Object> value = (Map<String, Object>) record.value();
```
mzitnik, line 40: Should we also check the type (String, byte[])  before casting I think we have Map only in the case of JSON deserializer

mzitnik:
Also seems like we already have 3 transformations we can publish a transformations bundle

mzitnik (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # 1.1.4
 * Bugfix to address field value to column name mapping for Tuples
+* Adding a KeyToValue transformation to allow for key to be stored in a separate column in ClickHouse
```
mshustov, line 3: @Paultagoras  it's new functionality, thus the version must be `1.2.0` to follow SemVer

PR #468, opened 2024-11-13
Get rid of the schema validation entirely.

Paultagoras:
So this ends up being the issue quite often, because things can be really finicky.


mzitnik:
I would prefer that we have a flag (if the user wants to disable it) since we do not know how much it protects or harm

PR #469, opened 2024-11-14
Trim excess fields

Paultagoras:
This should help reduce the payload size, if there is a lot of data being sent to Kafka but only a limited number of fields in the table itself.


mzitnik (approved):
Do we have a test? 
Going forward i would like us to convert it to RawBinary (and not send it as json - this should also take care of not sending unnecessary data)

PR #540, opened 2025-06-03
updated repo from graddle to maven and removed record data from logs 

pranav-g19:
## Checklist


mzitnik:
Sorry for the late response, can you please sign the CLA

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
```diff
@@ -109,7 +121,12 @@ private ClickHouseNode createClientV1() {
                 tmpJdbcConnectionProperties
         );
 
-        LOGGER.info("ClickHouse URL: {}", url);
+        LOGGER.info("[createClientV1] ClickHouse URL: {}", url);
+        LOGGER.info("[createClientV1] Hostname: {}", hostname);
```
mzitnik, line 125: We're logging hostname, port, and database, also in the previous section, and in the URL, do you think it is necessary to log it twice

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
```diff
@@ -141,9 +158,12 @@ private Client createClientV2() {
                 tmpJdbcConnectionProperties
         );
 
-        LOGGER.info("ClickHouse URL: {}", url);
-
-
+        LOGGER.info("[createClientV2] ClickHouse URL: {}", url);
+        LOGGER.info("[createClientV2] Hostname: {}", hostname);
```
mzitnik, line 162: Same here 
We're logging hostname, port, and database, also in the previous section, and in the URL, do you think it is necessary to log it twice

mzitnik (changes requested):
Can you share the reason for moving to Maven instead gradle, any reason why we are not using Maven wrapper https://maven.apache.org/wrapper/

mshustov:
as per the last message in Slack: ` Please ignore this PR.`

PR #565, opened 2025-07-31
[Performance] Added initial benchmark

chernser:
## Summary
Added initial project for the benchmark
## Checklist
image


laeg:
is this related to: https://github.com/ClickHouse/support-escalation/issues/5473

benchmark/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(17))
+    }
+}
+
+dependencies {
+    implementation("com.clickhouse.kafka:clickhouse-kafka-connect:v1.3.2")
```
mzitnik: Should we extract it as a variable?

benchmark/src/main/java/kafka_connector/data/SchemalessTestData.java
```diff
@@ -0,0 +1,366 @@
...
+                    System.currentTimeMillis(),
+                    TimestampType.CREATE_TIME
+            );
+
+            array.add(sr);
+        });
+        return array;
+    }
+
+    public static Collection<SinkRecord> createDataWithEmojis(String topic, int partition) {
```
mzitnik: Can we reuse it from our tests?

mzitnik (changes requested):
A few points 
- Can we convert it to be a module in the project so we could use the gradle directory 
- The code that creates data seems to identical to what we have in tests can we use it as a module

benchmark/build.gradle.kts
chernser: yes, this should be a variable.
btw in the latest gradle there is a designated file with versions.

benchmark/src/main/java/kafka_connector/data/SchemalessTestData.java
chernser: Yes, we just need to create a test jar and use it in the JMH project.

Issue #612, opened 2025-10-22
[monitoring] Implement ability to monitor number of inserted records and related metrics


chernser:
## Description 

Today connector do not report any metrics. There are logs but they are not convenient to collect and analyze. There should be another way of reporting telemetry. 

Most important metric we need is `number of inserted records` because it tells a real throughput of the connector. Another metric `time per batch` or similar would be also useful to detect slowness. 

Metrics: 
- `number of inserted records` 
- `bytes per batch` 
- `batch size` 
- `latency`

PR #631, opened 2025-11-08
[monitoring] Impl metrics

chernser:
## Summary

- Adds Sink tasks metrics 
- Adds topic metrics

Sink task and topic metrics are accessible via JMX. Bean name format is `com.clickhouse:type=ClickHouseKafkaConnector,name=SinkTask<taskId>,version=<version>`. 
For example, `com.clickhouse:type=ClickHouseKafkaConnector,name=SinkTask0,version=v1.3.5`.

Topic metrics bean name format is `com.clickhouse:type=ClickHouseKafkaConnector,name=SinkTask<taskId>,version=<version>,topic=<topic>`. 
For example `com.clickhouse:type=ClickHouseKafkaConnector,name=SinkTask0,version=v1.3.5,topic=schemaless_simple_batch_test_1762822854777`.

Closes https://github.com/ClickHouse/clickhouse-kafka-connect/issues/612

## Checklist

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/util/jmx/SimpleMovingAverage.java
+++ b/src/main/java/com/clickhouse/kafka/connect/util/jmx/SimpleMovingAverage.java
@@ -4,7 +4,11 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
 /**
- * Not thread-safe Exponential Moving Average implementation
+ * Not thread-safe Simple Moving Average implementation.
+ * This class accumulates sum of values and calculates average on demand.
+ * Values are stored in circular buffer to guarantee only last N values are used.
+ * It is needed to keep metric responsive after long periods of measurement.
+ *
  */
 public class SimpleMovingAverage {
 
@@ -22,6 +26,7 @@ public class SimpleMovingAverage {
 
     public void add(long value) {
         int insertIndex = head.getAndIncrement() % values.length;
+        // update sum by subtracting the oldest value (at insertIndex) and adding new value.
         sum.addAndGet(value - values[insertIndex]);
         values[insertIndex] = value;
     }
--- a/src/main/java/com/clickhouse/kafka/connect/util/jmx/SinkTaskStatisticsMBean.java
+++ b/src/main/java/com/clickhouse/kafka/connect/util/jmx/SinkTaskStatisticsMBean.java
@@ -62,7 +62,7 @@ public interface SinkTaskStatisticsMBean {
     /**
      * Mean receive lag calculated as receive by task time - record timestamp.
      * Calculated using a first record timestamp in the batch.
-     * Implemented as exponential moving average. Reset on restart.
+     * Implemented as simple moving average. Reset on restart.
      *
      * @return counter value
      */
--- a/src/main/java/com/clickhouse/kafka/connect/util/jmx/TopicStatisticsMBean.java
+++ b/src/main/java/com/clickhouse/kafka/connect/util/jmx/TopicStatisticsMBean.java
@@ -38,7 +38,7 @@ public interface TopicStatisticsMBean {
     /**
      * Mean receive lag what is calculated as receive by task time - record timestamp.
      * Calculated using a first record timestamp in the batch.
-     * Implemented as exponential moving average. Reset on restart.
+     * Implemented as simple moving average. Reset on restart.
      *
      * @return counter value
      */


build.gradle.kts
```diff
@@ -54,6 +54,7 @@ extra.apply {
     set("kafkaVersion", "2.7.0")
     set("gson", "2.13.1")
     set("jackson", "2.19.1")
+    set("dropwizardMetrics", "5.0.0")
```
mzitnik: Let's validate that Confluent is not using this package. We might need to shade it

src/main/java/com/clickhouse/kafka/connect/sink/processing/Processing.java
```diff
@@ -49,14 +51,23 @@ private void doInsert(List<Record> records, RangeContainer rangeContainer) {
             LOGGER.trace("doInsert - No records to insert.");
             return;
         }
-        QueryIdentifier queryId = new QueryIdentifier(records.get(0).getRecordOffsetContainer().getTopic(), records.get(0).getRecordOffsetContainer().getPartition(),
+
+        final Record firstRecord = records.get(0);
+        long eventReceiveLag = 0;
+        if (firstRecord.getSinkRecord().timestamp() != null) {
+            eventReceiveLag = System.currentTimeMillis() - firstRecord.getSinkRecord().timestamp();
```
mzitnik, line 58: Do we know where the timestamp comes from? Record metadata, or when created, just want to validate since we can have clock skew in some siuations

src/main/java/com/clickhouse/kafka/connect/util/jmx/SinkTaskStatistics.java
```diff
@@ -1,34 +1,168 @@
...
+        this.recordProcessingTime.addAndGet(timer.nanosElapsed());
     }
 
     public void taskProcessingTime(ExecutionTimer timer) {
-        this.taskProcessingTime += timer.nanosElapsed();
+        this.taskProcessingTime.addAndGet(timer.nanosElapsed());
     }
 
+
+    public void recordTopicStats(int n, String table, long eventReceiveLag) {
```
mzitnik, line 131: `int n` maybe `int numOfRecords`

build.gradle.kts
chernser: I've added this to the jar. 
I think it is not shared.

src/main/java/com/clickhouse/kafka/connect/sink/processing/Processing.java
chernser: I will find out. 
As for time skew - this would be a problem. But it can be normalized I think if skew stays in the range.

src/main/java/com/clickhouse/kafka/connect/util/jmx/SinkTaskStatistics.java
chernser: ok. will change

build.gradle.kts
chernser: This pulls too many dependencies. I will replace it
Thank you for pointing out!

src/main/java/com/clickhouse/kafka/connect/util/jmx/TopicStatistics.java
```diff
@@ -0,0 +1,68 @@
...
+    private final AtomicLong totalFailedRecords;
+
+    private final SimpleMovingAverage insertTime;
+
+    public TopicStatistics() {
+        totalInsertedRecords = new AtomicLong(0);
+        totalNumberOfBatches = new AtomicLong(0);
+        totalFailedBatches = new AtomicLong(0);
+        totalFailedRecords = new AtomicLong(0);
+        insertTime = new SimpleMovingAverage(SimpleMovingAverage.DEFAULT_WINDOW_SIZE);
```
mzitnik, line 19: Maybe we should expose it as a configuration?

src/main/java/com/clickhouse/kafka/connect/util/jmx/SimpleMovingAverage.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public double get() {
+       double sum = 0;
+       int count = head.get();
+       if (count >= values.length) {
+            count = values.length;
+       }
+
+       for (int i = 0; i < count; i++) {
```
mzitnik: We keep a sum 
Every time we add a new value, we add it to the sum and subtract the oldest from the array. 
This way, we can avoid having to sum up every time.

src/main/java/com/clickhouse/kafka/connect/util/jmx/TopicStatistics.java
chernser: I would like to keep it as is. 
I want to review monitoring and may be tune something - add another algo. for avg. values.

src/main/java/com/clickhouse/kafka/connect/util/jmx/SimpleMovingAverage.java
chernser: This is a good idea but need to determine what is the oldest in the array. 
I will work on it while improving statistics.

mzitnik: Just keep a pointer also to the oldest one

chernser: actually I will just subtract value of the element where I'm going to write a new value - it will be the oldest value in circular array.

src/main/java/com/clickhouse/kafka/connect/util/jmx/SimpleMovingAverage.java
```diff
@@ -0,0 +1,34 @@
...
+ * Not thread-safe Exponential Moving Average implementation
+ */
+public class SimpleMovingAverage {
+
+    public static final int DEFAULT_WINDOW_SIZE = 60;
+
+    private final long[] values;
+    private final AtomicInteger head;
+    private final AtomicLong sum;
+    private final int n;
```
mzitnik: Where is n used?

src/main/java/com/clickhouse/kafka/connect/util/jmx/SimpleMovingAverage.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public SimpleMovingAverage(int numOfValues) {
+        this.values = new long[numOfValues];
+        this.n = this.values.length - 1;
+        this.head = new AtomicInteger();
+        this.sum = new AtomicLong();
+    }
+
+    public void add(long value) {
```
mzitnik, line 27: Can you explain the logic here

src/main/java/com/clickhouse/kafka/connect/util/jmx/SimpleMovingAverage.java
chernser: holds values.length - 1 value. just not to do it all the time.

src/main/java/com/clickhouse/kafka/connect/util/jmx/SimpleMovingAverage.java
chernser: getting index where to write a new value 
subtract existing value - because will override it 
add new value 
write value to the array. 

It is what I think you are suggesting. The oldest element is that we are going to replace.

mzitnik: Yes it is - just let's explain the algoritim in source code (for future referance)

src/main/java/com/clickhouse/kafka/connect/util/jmx/SimpleMovingAverage.java
mzitnik: We are setting it once and never using it?

PR #645, opened 2025-11-26
Lowered log message level from error to warn as it is not an error

chernser:
## Summary
There is a message `Got non-root column, but its parent was not found to be updated.` when processing ephemeral columns. This happens because only `.null` is present for the column but not itself. This doesn't affect processing but floods logs. Thus we lower log level for this message and will revisit this part of code later.  

## Checklist


mzitnik (approved):
LGTM

PR #658, opened 2026-02-13
[FEAT] buffer implementation

gaurav7261:
## Summary
• Buffer implementation: Adds internal record buffering to ClickHouseSinkTask so that records from multiple poll() calls are accumulated and flushed as a single large batch, reducing the number of
    inserts to ClickHouse.
  • Two new config options: bufferCount (number of records to accumulate before flushing) and bufferFlushTime (max time in ms before flushing regardless of count). Both default to 0 (disabled), preserving
    backward compatibility.

## Checklist

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
+++ b/src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
@@ -11,7 +11,9 @@ import org.apache.kafka.connect.sink.SinkTask;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 
 public class ClickHouseSinkTask extends SinkTask {
@@ -21,6 +23,13 @@ public class ClickHouseSinkTask extends SinkTask {
     private ProxySinkTask proxySinkTask;
     private ClickHouseSinkConfig clickHouseSinkConfig;
     private ErrorReporter errorReporter;
+
+    // Internal buffering
+    private List<SinkRecord> buffer;
+    private long lastFlushTime;
+    private int bufferCount;
+    private long bufferFlushTime;
+    private boolean bufferingEnabled;
 
     @Override
     public String version() {
@@ -40,11 +49,59 @@ public class ClickHouseSinkTask extends SinkTask {
         }
 
         this.proxySinkTask = new ProxySinkTask(clickHouseSinkConfig, errorReporter);
+
+        // Initialize buffering
+        this.bufferCount = clickHouseSinkConfig.getBufferCount();
+        this.bufferFlushTime = clickHouseSinkConfig.getBufferFlushTime();
+        this.bufferingEnabled = this.bufferCount > 0;
+        this.buffer = new ArrayList<>();
+        this.lastFlushTime = System.currentTimeMillis();
+
+        if (bufferingEnabled) {
+            LOGGER.info("Internal buffering enabled: bufferCount={}, bufferFlushTime={}ms",
+                    bufferCount, bufferFlushTime);
+        }
     }
 
 
     @Override
     public void put(Collection<SinkRecord> records) {
+        if (!bufferingEnabled) {
+            // Original behavior - no buffering
+            putDirect(records);
+            return;
+        }
+
+        // Buffering mode: accumulate records
+        if (!records.isEmpty()) {
+            buffer.addAll(records);
+            LOGGER.debug("Buffered {} records, total buffer size: {}", records.size(), buffer.size());
+        }
+
+        // Check if we should flush
+        boolean sizeThreshold = buffer.size() >= bufferCount;
+        boolean timeThreshold = bufferFlushTime > 0
+                && (System.currentTimeMillis() - lastFlushTime) >= bufferFlushTime
+                && !buffer.isEmpty();
+
+        if (sizeThreshold || timeThreshold) {
+            LOGGER.info("Buffer flush triggered: size={}, sizeThreshold={}, timeThreshold={}", buffer.size(), sizeThreshold, timeThreshold);
+            flushBuffer();
+        }
+    }
+
+    private void flushBuffer() {
+        if (buffer.isEmpty()) {
+            return;
+        }
+        List<SinkRecord> toFlush = new ArrayList<>(buffer);
+        putDirect(toFlush);
+        // Clear buffer only after successful flush to prevent data loss on failure
+        buffer.clear();
+        lastFlushTime = System.currentTimeMillis();
+    }
+
+    private void putDirect(Collection<SinkRecord> records) {
         try {
             long putStat = System.currentTimeMillis();
             this.proxySinkTask.put(records);
@@ -63,7 +120,6 @@ public class ClickHouseSinkTask extends SinkTask {
         }
     }
 
-
     // TODO: can be removed ss
     @Override
     public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
@@ -72,6 +128,11 @@ public class ClickHouseSinkTask extends SinkTask {
 
     @Override
     public void stop() {
+        // Flush remaining buffered records before stopping
+        if (bufferingEnabled && buffer != null && !buffer.isEmpty()) {
+            LOGGER.info("Stop called with {} buffered records - flushing", buffer.size());
+            flushBuffer();
+        }
         if (this.proxySinkTask != null) {
             this.proxySinkTask.stop();
         }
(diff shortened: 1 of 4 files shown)


src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -40,11 +51,74 @@ public void start(Map<String, String> props) {
         }
 
         this.proxySinkTask = new ProxySinkTask(clickHouseSinkConfig, errorReporter);
+
+        // Initialize buffering
+        this.bufferCount = clickHouseSinkConfig.getBufferCount();
+        this.bufferFlushTime = clickHouseSinkConfig.getBufferFlushTime();
+        this.bufferingEnabled = this.bufferCount > 0;
+        this.buffer = new ArrayList<>();
```
mzitnik: should we realocate in advance by using the BUFFER_COUNT

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -40,11 +51,74 @@ public void start(Map<String, String> props) {
...
+        this.lastFlushTime = System.currentTimeMillis();
+        this.flushedOffsets = new HashMap<>();
+
+        if (bufferingEnabled && clickHouseSinkConfig.isExactlyOnce()) {
+            throw new ConnectException("Internal buffering (bufferCount > 0) is not compatible with exactly-once mode. " +
```
mzitnik:
```suggestion
            throw new ConnectException(
    "Internal buffering (bufferCount > 0) is not compatible with exactly-once mode. " +
    "Buffering changes batch boundaries, which breaks ClickHouse block deduplication and the offset state machine. " +
    "To resolve this, either disable exactly-once mode by setting 'exactlyOnce =false' in your connector config, " +
    "or disable buffering by setting 'bufferCount=0'."
);
```

mquatrini:
We're running clickhouse-kafka-connect v1.3.5 in production and have identified that the lack of internal buffering causes severe throughput degradation at low traffic — the connector makes one ClickHouse INSERT per partition per poll cycle (we have ~45 partitions per task), resulting in ~750 rec/sec total throughput overnight when it needs to be ~50K rec/sec to recover from peak-hour backlog. This PR directly addresses our root cause. @mzitnik could you prioritize this review?

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -39,12 +50,80 @@ public void start(Map<String, String> props) {
...
+        }
+
+        // Check if we should flush
+        boolean sizeThreshold = buffer.size() >= bufferCount;
+        boolean timeThreshold = bufferFlushTime > 0
+                && (System.currentTimeMillis() - lastFlushTime) >= bufferFlushTime
+                && !buffer.isEmpty();
+
+        if (sizeThreshold || timeThreshold) {
+            LOGGER.info("Buffer flush triggered: size={}, sizeThreshold={}, timeThreshold={}", buffer.size(), sizeThreshold, timeThreshold);
```
chernser: Please make a debug level because will be very noisy on production. and needed only for tuning.

chernser: additionally may be  add last `flushTime`

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -63,15 +142,59 @@ public void put(Collection<SinkRecord> records) {
...
+        // Records still in the buffer have NOT been written, so their offsets must not be committed.
+        // This guarantees at-least-once delivery: on crash/rebalance, Kafka redelivers buffered records.
+        // Pattern follows Confluent S3 Sink Connector: flush() is no-op, preCommit() manages offsets.
+        if (flushedOffsets.isEmpty()) {
+            LOGGER.info("preCommit: no offsets to commit (all records still buffered)");
+            return new HashMap<>();
+        }
+        Map<TopicPartition, OffsetAndMetadata> offsetsToCommit = new HashMap<>(flushedOffsets);
+        flushedOffsets.clear();
+        LOGGER.info("preCommit: committing offsets for flushed data: {}", offsetsToCommit);
```
chernser: please make a debug level  because will be too many records in the log.

chernser (changes requested):

chernser (approved):

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -63,15 +142,59 @@ public void put(Collection<SinkRecord> records) {
...
 
     @Override
     public void stop() {
+        // Note: close() is called before stop() by the framework, which already
+        // removes buffered records. Unflushed records' offsets were never committed
+        // via preCommit(), so Kafka will redeliver them on restart (at-least-once).
+        if (bufferingEnabled && buffer != null && !buffer.isEmpty()) {
+            LOGGER.warn("Stop called with {} buffered records still in buffer — " +
+                    "these will be redelivered on restart since offsets were not committed", buffer.size());
+            buffer.clear();
```
suryakantade, line 196: Since stop() represents a full task shutdown (not just partition revocation),  should we attempt to flush the buffer before clearing it?

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -39,12 +50,80 @@ public void start(Map<String, String> props) {
...
+
+        // Check if we should flush
+        boolean sizeThreshold = buffer.size() >= bufferCount;
+        boolean timeThreshold = bufferFlushTime > 0
+                && (System.currentTimeMillis() - lastFlushTime) >= bufferFlushTime
+                && !buffer.isEmpty();
+
+        if (sizeThreshold || timeThreshold) {
+            LOGGER.debug("Buffer flush triggered: size={}, sizeThreshold={}, timeThreshold={}, lastFlushTime={}", buffer.size(), sizeThreshold, timeThreshold, lastFlushTime);
+            flushBuffer();
```
suryakantade, line 102: max.poll.interval.ms must be larger than your worst-case flushBuffer() latency.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -63,15 +142,59 @@ public void put(Collection<SinkRecord> records) {
...
+        if (!bufferingEnabled) {
+            return;
+        }
+        // Remove buffered records for revoked partitions to prevent duplicates.
+        // Their offsets were never committed via preCommit(), so the new owner
+        // will redeliver them — no data loss.
+        if (!buffer.isEmpty()) {
+            int before = buffer.size();
+            buffer.removeIf(record -> {
+                TopicPartition tp = new TopicPartition(record.topic(), record.kafkaPartition());
```
suryakantade, line 175: It would be incompatible with Kafka's upcoming Share Groups ([KIP-932](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka#KIP932:QueuesforKafka-Status)), which use record-level acknowledgments and allow multiple consumers per partition.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: We intentionally skip flushing here — _close()_ is already called before _stop()_ by the framework, removing buffered records for revoked partitions. More importantly, since _preCommit()_ only  commits offsets for data successfully written to ClickHouse, any buffered but unflushed records never had their offsets committed. Kafka will redeliver them on restart, preserving at-least-once semantics.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: valid point, we faced this when we not received ack from one of ch shard, and fix is here: https://github.com/ClickHouse/clickhouse-kafka-connect/pull/756

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: yes, Share Groups compatibility would require a fundamentally different acknowledgment strategy (per-record rather than per-partition offset tracking). Agreed this should be documented as a limitation. Out of scope for this PR but worth tracking as a future issue.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
suryakantade: Awesome !!

Issue #646, opened 2025-12-09
Integration testing


chernser:
## Description 

Implement running integration test

PR #665, opened 2026-03-04
fix integration tests

kurnoolsaketh:
## Summary
This PR fixes the integration tests and reorganizes/reformats the code in the test files. Brief change summary by file:
- ClickHouseSinkConnectorIntegrationTest.java: transparently connect to the toxiproxy endpoint and avoid configuring the proxy settings on the V1 client. The proxy will relay relative URIs, which the CH server expects.
- ClickHouseCloudTest.java: remove the `indexCount` column and use `off16` to check the sequential row count instead. Previously, `indexCount` was not being written and defaulted to `0` in the CH table. This resulted in assertion failure when comparing the actual number of distinct values (1) to the expected number (1000). `SELECT DISTINCT `off16`` produces the desired behavior.
- ExactlyOnceTest.java: minor code refactoring - these tests were not broken.
- .github/integrationTests.yaml: Previously, the arch was hardcoded to x64, which is incompatible with the `cnfldemos/cp-server-connect-datagen` [docker image](https://hub.docker.com/r/cnfldemos/cp-server-connect-datagen/tags). Change the platform to arm64/aarch64 instead.

closes https://github.com/ClickHouse/clickhouse-kafka-connect/issues/646

## Validation
clean run here: https://github.com/ClickHouse/clickhouse-kafka-connect/actions/runs/22695986263


src/integrationTest/java/com/clickhouse/kafka/connect/sink/ClickHouseCloudTest.java
```diff
@@ -102,8 +108,35 @@ public void overlappingDataTest() {
...
         LOGGER.info("Total Records: {}", sr.size());
         LOGGER.info("Row Count: {}", ClickHouseTestHelpers.countRows(chc, topic));
-        assertTrue(ClickHouseTestHelpers.countRows(chc, topic) >= sr.size());
-        assertTrue(ClickHouseTestHelpers.checkSequentialRows(chc, topic, sr.size()));
+        Assertions.assertTrue(ClickHouseTestHelpers.countRows(chc, topic) >= sr.size());
+        Assertions.assertTrue(checkSequentialRows(chc, topic, sr.size()));
         ClickHouseTestHelpers.dropTable(chc, topic);
     }
+
+    private static boolean checkSequentialRows(ClickHouseHelperClient chc, String tableName, int totalRecords) {
```
mzitnik: why did we moved it from ClickHouseTestHelpers

.github/workflows/integrationTests.yaml
```diff
@@ -4,7 +4,7 @@ on: [workflow_dispatch]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
```
chernser, line 10: Is it for purpose?

chernser: I read the comment about image limitations. question is why arm ?

src/integrationTest/java/com/clickhouse/kafka/connect/sink/ClickHouseCloudTest.java
kurnoolsaketh: discussed offline, decided to move back to the helper class

.github/workflows/integrationTests.yaml
kurnoolsaketh: it's either arm or amd - i confirmed all docker images that the tests use are compatible with arm, so that's the one i chose. Any concern you see?

chernser: Just curious why do we need `arm`  when image requires `amd64` 
https://hub.docker.com/r/cnfldemos/cp-server-connect-datagen/tags

Image

kurnoolsaketh: the test is using v0.6.2-7.5.0

https://hub.docker.com/layers/cnfldemos/cp-server-connect-datagen/0.6.2-7.5.0/images/sha256-64797080ff2fad7e01b5cf017b2e30fd9aa90770d9829c5cf1785ed606d0d80d

kurnoolsaketh: it requires either amd or arm

chernser: yes, AMD = x64 (AMD introduced 64 bit)  just naming I think. 
Here is runners list https://docs.github.com/en/actions/reference/runners/github-hosted-runners

But I'm fine - lets discuss offline.

CONTRIBUTING.md
```diff
@@ -1,14 +1,22 @@
...
 * A local copy of the repo, pulled from git
 
 ## Building and Running Unit Tests
-You should be able to compile + run the unit tests locally by going to the root project folder and running `./gradlew clean test`. Note this doesn't produce a release artifact, you'll have to execute a later step for that.
+You can compile and run the unit tests locally by going to the root project folder and running `./gradlew clean test`. _Note: this doesn't produce a release artifact, you'll have to execute a later step for that._
+
+## Building and Running Integration Tests
+You can compile and run the integration test suite locally by going to the root project folder and running `./gradlew clean integrationTest`. The integration tests also run nightly as a GitHub action.
+
+_Note: `ClickHouseCloudTest` and `ExactlyOnceTest` run against cloud instances only - they will automatically be skipped if the required system options are not set, specifically:_
```
mzitnik, line 16: Looking at our `integrationTests.yaml`, we also ask to set clickhouse.cloud.organization, clickhouse.cloud.id, clickhouse.cloud.secret, clickhouse.cloud.serviceId, is this mandatory to run tests against the cloud?

mzitnik: Let's also please explain how to run it 
`./gradlew clean integrationTest -Dclickhouse.host=....`

mzitnik (approved):
I'am approving, but lets explan how to run it.

PR #676, opened 2026-03-12
Replace Gson with Jackson ObjectMapper for JSON insert serialization

gaurav7261:
## Summary
Switch doInsertJsonV1 and doInsertJsonV2 from per-call Gson instances to a shared static Jackson ObjectMapper. Jackson's writeValueAsBytes() serializes directly to UTF-8 bytes, eliminating the
  intermediate UTF-16 String allocation that Gson's toJson() + getBytes() required — reducing memory pressure and GC churn on the hot insert path.
  Also:
  • Register a custom JsonSerializer<Struct> so Jackson can serialize Kafka Connect Struct objects as clean {field: value} JSON (replaces Gson's reflection-based internal representation that exposed
    schema and values internals)
  • Guard LOGGER.trace with isTraceEnabled() to avoid constructing payload strings when trace is disabled
  • Use ((Number) value).shortValue/intValue/longValue for UINT casts to prevent ClassCastException when numeric types don't exactly match (e.g., Long value for UInt16 column)
  • Remove unused Gson imports and dead SchemaFieldExclusionStrategy inner class
  • Update FailureTest to use STRING_SCHEMA for genuinely incompatible type mismatch (previously relied on ClassCastException from strict casts)

  **Performance Impact: Gson to Jackson Migration (production measurements)**
  By replacing Gson with Jackson's ObjectMapper.writeValueAsBytes(), we eliminate the intermediate UTF-16 String allocation that Gson required (Object → String → bytes), letting Jackson serialize
  directly to UTF-8 bytes (Object → bytes). This cuts serialization time nearly in half.
  • Serialization dropped from ~5.0s to ~2.7s per batch (45% faster) — direct win from Jackson's writeValueAsBytes() skipping the UTF-16 to UTF-8 transcoding step and avoiding per-call new Gson() +
    TypeToken allocations.
  • Total data prep dropped from ~5.3s to ~2.8s (47% faster) — the serialization savings flow through to the overall data preparation phase, which includes field cleanup and byte writing.

## Checklist

### DIFF
--- a/src/test/java/com/clickhouse/kafka/connect/sink/db/JsonSerializationEquivalenceTest.java
+++ b/src/test/java/com/clickhouse/kafka/connect/sink/db/JsonSerializationEquivalenceTest.java
@@ -1,5 +1,7 @@
 package com.clickhouse.kafka.connect.sink.db;
 
+import com.clickhouse.kafka.connect.util.DataJson;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
@@ -20,7 +22,8 @@ import static org.junit.jupiter.api.Assertions.*;
 
 /**
  * Verifies that the Gson→Jackson migration produces byte-equivalent output
- * for all Map shapes used in the JSON insert paths (doInsertJsonV1/V2).
+ * for all Map shapes used in the JSON insert paths (doInsertJsonV1/V2),
+ * and semantically equivalent output for Struct serialization.
  *
  * @author Gaurav Miglani
  */
@@ -29,6 +32,28 @@ public class JsonSerializationEquivalenceTest {
     private static final Gson GSON = new Gson();
     private static final java.lang.reflect.Type GSON_MAP_TYPE =
             new TypeToken<HashMap<String, Object>>() {}.getType();
+
+    /** DataJson.GSON has the StructSerializer registered (via StructToJsonMap). */
+    private static final Gson STRUCT_GSON = DataJson.GSON;
+
+    /** Shared reader for JSON tree comparisons. */
+    private static final ObjectMapper JSON_READER = new ObjectMapper();
+
+    /**
+     * Asserts that two JSON byte arrays are semantically equivalent (same
+     * key-value pairs regardless of field order).  Gson's StructSerializer
+     * converts Struct fields via StructToJsonMap into a HashMap whose
+     * iteration order is undefined, while Jackson iterates in schema-defined
+     * order.  The JSON is semantically identical — ClickHouse JSONEachRow
+     * matches by field name, not position.
+     */
+    private static void assertJsonEquivalent(byte[] gsonBytes, byte[] jacksonBytes, String message) throws Exception {
+        JsonNode gsonTree = JSON_READER.readTree(gsonBytes);
+        JsonNode jacksonTree = JSON_READER.readTree(jacksonBytes);
+        assertEquals(gsonTree, jacksonTree, message);
+    }
+
+    // ── Map equivalence tests (byte-identical) ──────────────────────────
 
     @Test
     void flatMapWithPrimitives() throws Exception {
@@ -133,6 +158,8 @@ public class JsonSerializationEquivalenceTest {
                 "Map with special characters should produce identical bytes");
     }
 
+    // ── Struct equivalence tests (Gson DataJson.GSON vs Jackson) ────────
+
     @Test
     void structSerializesToCleanJson() throws Exception {
         Schema schema = SchemaBuilder.struct()
@@ -146,10 +173,13 @@ public class JsonSerializationEquivalenceTest {
                 .put("age", 25)
                 .put("active", false);
 
+        byte[] gsonBytes = STRUCT_GSON.toJson(struct).getBytes(StandardCharsets.UTF_8);
         byte[] jacksonBytes = OBJECT_MAPPER.writeValueAsBytes(struct);
 
-        ObjectMapper reader = new ObjectMapper();
-        Map<?, ?> parsed = reader.readValue(jacksonBytes, Map.class);
+        assertJsonEquivalent(gsonBytes, jacksonBytes,
+                "Flat struct: Gson StructSerializer and Jackson JacksonStructSerializer should produce equivalent JSON");
+
+        Map<?, ?> parsed = JSON_READER.readValue(jacksonBytes, Map.class);
         assertEquals("Bob", parsed.get("name"));
         assertEquals(25, parsed.get("age"));
         assertEquals(false, parsed.get("active"));
@@ -175,10 +205,13 @@ public class JsonSerializationEquivalenceTest {
                 .put("name", "Charlie")
                 .put("address", address);
 
+        byte[] gsonBytes = STRUCT_GSON.toJson(person).getBytes(StandardCharsets.UTF_8);
         byte[] jacksonBytes = OBJECT_MAPPER.writeValueAsBytes(person);
-        ObjectMapper reader = new ObjectMapper();
-        Map<?, ?> parsed = reader.readValue(jacksonBytes, Map.class);
 
+        assertJsonEquivalent(gsonBytes, jacksonBytes,
+                "Nested struct: Gson and Jackson should produce equivalent JSON");
+
+        Map<?, ?> parsed = JSON_READER.readValue(jacksonBytes, Map.class);
         assertEquals("Charlie", parsed.get("name"));
         @SuppressWarnings("unchecked")
         Map<String, Object> addr = (Map<String, Object>) parsed.get("address");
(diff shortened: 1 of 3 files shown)


kurnoolsaketh:
integration test run here: https://github.com/ClickHouse/clickhouse-kafka-connect/actions/runs/23094463586

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
kurnoolsaketh: GSON is still being used in the raw binary insertion codepath (follow the `doInsertRawBinary` call from line 236 to get here) - can you replace this with OM as well and remove the gson import from this file?

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -1356,19 +1380,4 @@ public long recordsInserted() {
         return 0;
     }
 
-    private static class SchemaFieldExclusionStrategy implements ExclusionStrategy {
```
kurnoolsaketh, line 1377: why remove this?

src/test/java/com/clickhouse/kafka/connect/sink/dlq/FailureTest.java
```diff
@@ -85,7 +85,7 @@ public static List<SinkRecord> createInvalidRecords(String topic, int partition,
         Schema NESTED_SCHEMA = SchemaBuilder.struct()
                 .field("off16", Schema.INT16_SCHEMA)
                 .field("uint8", Schema.OPTIONAL_INT8_SCHEMA)
-                .field("uint16", Schema.INT64_SCHEMA)
+                .field("uint16", Schema.STRING_SCHEMA)
```
kurnoolsaketh, line 88: @chernser please review the change to this test

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -72,6 +73,30 @@
...
+                    } else {
+                        provider.defaultSerializeValue(value, gen);
+                    }
+                }
+                gen.writeEndObject();
+            }
+        });
+        mapper.registerModule(module);
+        return mapper;
+    }
```
kurnoolsaketh: a better place for this is DataJson.java - this way the static initializer can be called once and used everywhere (the same way GSON is right now)

kurnoolsaketh (commented):
@gaurav7261 thanks for the change - made a few comments. 

Also, can you add a unit test to cover the migration? We need to be absolutely sure `GSON.toJson(...).getBytes(StandardCharsets.UTF_8)` produces the `.equals()` byte string as `OBJECT_MAPPER.writeValueAsBytes(...)` everywhere it is used.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
gaurav7261: @kurnoolsaketh  `SchemaFieldExclusionStrategy` was dead code — it was never registered with any Gson instance. The local new `Gson()` calls in doInsertJsonV1/V2 used a plain default Gson (no exclusion strategies), and the
     shared DataJson.GSON uses its own DataSerializer/StructSerializer instead. Removing it as part of the Gson cleanup

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
gaurav7261: done, please review @kurnoolsaketh

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
gaurav7261: done, please review @kurnoolsaketh

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -1020,15 +1044,17 @@ protected void doInsertJsonV1(List<Record> records, Table table, QueryIdentifier
                                 break;
                         }
                         long beforeSerialize = System.currentTimeMillis();
-                        String gsonString = gson.toJson(cleanupExtraFields(data, table), gsonType);
+                        Map<String, Object> cleaned = cleanupExtraFields(data, table);
+                        byte[] bytes = OBJECT_MAPPER.writeValueAsBytes(cleaned);
                         dataSerializeTime += System.currentTimeMillis() - beforeSerialize;
```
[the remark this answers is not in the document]
gaurav7261: i think this is handled in integration test link you have pasted correct @kurnoolsaketh ?

src/test/java/com/clickhouse/kafka/connect/sink/dlq/FailureTest.java
mzitnik: @gaurav7261 Can you explain?

gaurav7261: It used Schema.INT64_SCHEMA with a valid numeric value `((long) ThreadLocalRandom.current().nextInt(0, 32767))`. After the Gson-to-Jackson migration, Jackson's serializer became more lenient with numeric type coercion — it can successfully cast an INT64 value into a UInt16 column without error. This meant the "invalid" records were actually being inserted successfully, causing the DLQ test to fail because no records were being rejected.

src/test/java/com/clickhouse/kafka/connect/sink/db/JsonSerializationEquivalenceTest.java
```diff
@@ -0,0 +1,358 @@
...
+        Map<String, Object> map = new LinkedHashMap<>();
+        map.put("name", "Alice");
+        map.put("optional", null);
+
+        String gsonJson = GSON.toJson(map, GSON_MAP_TYPE);
+        String jacksonJson = new String(OBJECT_MAPPER.writeValueAsBytes(map), StandardCharsets.UTF_8);
+
+        assertFalse(gsonJson.contains("optional"),
+                "Gson drops null values from maps");
+        assertTrue(jacksonJson.contains("\"optional\":null"),
```
chernser: this is a behavior change. I think, we need to make jackson to ignore null fields. 
CC: @mzitnik

gaurav7261: Gson silently drops null values from Maps, which means ClickHouse would use column DEFAULTs for those fields. Jackson was preserving them as explicit null, which could override custom non-null DEFAULTs on Nullable columns. Fixed by configuring the ObjectMapper with `JsonInclude.Include.NON_NULL`. This only affects Map serialization — our custom Struct serializer still writes nulls explicitly (for the binary format JSON column path), matching the original Gson Struct behavior, now making sense @chernser @mzitnik ?

src/test/java/com/clickhouse/kafka/connect/sink/db/JsonSerializationEquivalenceTest.java
```diff
@@ -0,0 +1,359 @@
...
+                .field("age", Schema.INT32_SCHEMA)
+                .field("active", Schema.BOOLEAN_SCHEMA)
+                .build();
+
+        Struct struct = new Struct(schema)
+                .put("name", "Bob")
+                .put("age", 25)
+                .put("active", false);
+
+        byte[] jacksonBytes = OBJECT_MAPPER.writeValueAsBytes(struct);
```
kurnoolsaketh, line 177: can you assert that this and the GSON bytes are the same? If they are not the same, do you know what the expected behavior is when the struct is serialized using Jackson and written to CH?

kurnoolsaketh: same comment applies to all Struct tests below

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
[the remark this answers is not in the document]
kurnoolsaketh: no, i don't believe the integration tests explicitly validate struct serialization. I believe the unit tests you added in `JsonSerializationEquivalenceTest.java` can be adapted into an integration test (encapsulate the `Struct`s in a `List<Record>` and send them to CH) - are you able to do this?

src/test/java/com/clickhouse/kafka/connect/sink/db/JsonSerializationEquivalenceTest.java
kurnoolsaketh: alternatively, I would feel ok if this were an integration test instead. See my comment [here](https://github.com/ClickHouse/clickhouse-kafka-connect/pull/676/changes#r2962076600).

gaurav7261: done @kurnoolsaketh ,please review

PR #687, opened 2026-03-23
Feature/schema evolution

### LINKED ISSUE #277: Support automatic table schema evolution
**Description**

In my organization we use clickhouse as analitical database that gets it's data from multiple sources via Kafka. Domain models change from time to time as business changes and we found a need to reflect those changes in the clickhouse manually a little inconvenient.

Would be awesome if this connector could propagate changes in schemas to the clickhouse tables automatically. 

**Additional context**
Compatibility modes in kafka:   
https://docs.confluent.io/cloud/current/sr/fundamentals/schema-evolution.html#compatibility-types

JDBC connector that has this functionality:
https://github.com/confluentinc/kafka-connect-jdbc

guilleov:
# Implement `auto.evolve` schema evolution (closes #277)

## Summary

Adds automatic table schema evolution to the ClickHouse Kafka Connect sink connector. When `auto.evolve=true`, the connector detects new fields in incoming Kafka records and issues `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` statements against ClickHouse - no manual DDL required when upstream schemas change.

This mirrors the [auto.evolve feature](https://docs.confluent.io/kafka-connectors/jdbc/current/sink-connector/sink_config_options.html#auto-evolution) in the Confluent JDBC Sink Connector, adapted for ClickHouse.

**Resolves:** [#277 - Support automatic table schema evolution](https://github.com/ClickHouse/clickhouse-kafka-connect/issues/277)

## Why this feature

When a Kafka topic's schema changes (e.g. a new field is added to an Avro/Protobuf/JSON Schema definition), the connector currently either silently drops the new field (`input_format_skip_unknown_fields=1`) or fails. Users must manually issue `ALTER TABLE` statements to add columns before the connector can ingest the new fields. This is inconvenient and error-prone, especially in organizations with frequent domain model changes ([original issue context](https://github.com/ClickHouse/clickhouse-kafka-connect/issues/277)).

### Idempotent DDL

[Issue #277 raised a concern](https://github.com/ClickHouse/clickhouse-kafka-connect/issues/277#issuecomment-2149453291) about race conditions when multiple connector tasks run concurrently. ClickHouse has `ADD COLUMN IF NOT EXISTS`, which makes the DDL **idempotent** - if two tasks race to add the same column, both will succeed.

### Async DDL propagation on replicated tables

For replicated tables, ALTER queries add instructions to ZooKeeper/Keeper and are applied [asynchronously on replicas](https://clickhouse.com/docs/sql-reference/statements/alter#synchronicity-of-alter-queries). Two mitigations:

1. **`alter_sync=1` on DDL statements** - Waits for the local replica to apply the change before returning. This eliminates most staleness without the risk of `alter_sync=2` (which blocks until ALL replicas apply and can hang if a replica is down).
2. **Retry loop on DESCRIBE TABLE** - After DDL, `refreshTableAfterDDL()` verifies the expected new columns are visible. If not (e.g., reading from a different replica), it retries up to `auto.evolve.ddl.refresh.retries` times (default 3) with 200ms backoff.

### Type mapping

- **Primitive types** (Int, Float, String, Bool, etc.) are wrapped in `Nullable(...)` so older records missing the field get NULL.
- **Array and Map types** cannot be Nullable in ClickHouse, so they are created with an explicit `DEFAULT` expression (`DEFAULT []` for Array, `DEFAULT map()` for Map). This allows `RowBinaryWithDefaults` to emit the "use server default" marker for older records that lack the field.
- **Variant types** (from Avro/Protobuf union schemas with distinct branches) are created without Nullable or DEFAULT - Variant has an implicit NULL discriminator.
- **Struct types** can optionally be mapped to ClickHouse `JSON` columns via `auto.evolve.struct.to.json=true`.
- **Logical types** (Decimal, Date, Timestamp, Time) are mapped to their ClickHouse equivalents (Decimal128, Date32, DateTime64, Int64).

### Configuration

| Property | Type | Default | Description |
|---|---|---|---|
| `auto.evolve` | boolean | `false` | Enable automatic schema evolution |
| `auto.evolve.ddl.refresh.retries` | int | `3` | Retries when waiting for DDL propagation |
| `auto.evolve.struct.to.json` | boolean | `false` | Map STRUCT fields to ClickHouse JSON columns |

## Backward compatibility

- `auto.evolve` defaults to `false` - existing deployments are completely unaffected
- No record processing logic changes behavior for users who don't enable the flag

## Tests

30 new auto-evolve tests in `ClickHouseSinkTaskWithSchemaTest.java` covering:
- Basic column addition (nullable, non-nullable, multiple columns)
- Mixed schema batches (V1+V2, V1+V2+V3, interleaved versions, cross-partition)
- Array, Map, and typed Array column evolution
- Variant/union column evolution
- Struct-to-JSON mapping
- Logical types (Date, Timestamp, Decimal)
- DDL refresh timeout and failure handling
- Schemaless record rejection
- Cache behavior after DDL

**Full test suite:** 264 tests, 261 succeeded, 0 failed, 3 skipped

## Checklist

- Unit and integration tests covering the common scenarios were added
- Full test suite passes (`./gradlew test` - 0 failures)
- A human-readable description of the changes was provided to include in CHANGELOG


src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
```diff
@@ -470,6 +470,43 @@ public Table describeTableV2(String database, String tableName) {
...
+        } catch (ClickHouseException e) {
+            throw new RuntimeException("Failed to execute ALTER TABLE: " + sql, e);
+        }
+    }
+
+    private void alterTableAddColumnV2(String sql) {
+        try {
+            QuerySettings settings = new QuerySettings();
+            settings.serverSetting("alter_sync", "1");
+            client.query(sql, settings).get();
```
chernser: it returns response object that should be closed to not leak connection.

src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Column.java
```diff
@@ -358,6 +363,71 @@ private static Map<String, Integer> extractEnumValues(String valueType) {
         return data;
     }
 
+    public static String connectTypeToClickHouseType(Schema connectSchema) {
+        // Check logical types first (same pattern as JDBC connector)
+        if (connectSchema.name() != null) {
+            switch (connectSchema.name()) {
+                case Decimal.LOGICAL_NAME:
+                    int precision = 38; // ClickHouse Decimal128 default
```
chernser: please extract to a static field - it will be more clear without a comment.

src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Column.java
```diff
@@ -358,6 +363,71 @@ private static Map<String, Integer> extractEnumValues(String valueType) {
...
+                return "Array(" + elementType + ")";
+            case MAP:
+                String keyType = connectTypeToClickHouseType(connectSchema.keySchema());
+                String valType = connectTypeToClickHouseType(connectSchema.valueSchema());
+                if (connectSchema.valueSchema().isOptional()) {
+                    valType = "Nullable(" + valType + ")";
+                }
+                return "Map(" + keyType + ", " + valType + ")";
+            case STRUCT:
+                throw new RuntimeException(
```
chernser: here we can add one more flag to allow creating JSON for struct 
Also we need to handle unions because it is common case.

src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Column.java
```diff
@@ -358,6 +363,71 @@ private static Map<String, Integer> extractEnumValues(String valueType) {
...
+                        "Cannot auto-evolve STRUCT fields to ClickHouse columns. " +
+                        "STRUCT type requires manual mapping to Tuple, JSON, or Nested type.");
+            default:
+                throw new RuntimeException("Unsupported Connect type for auto-evolution: " + connectSchema.type());
```
chernser: Let's create a `SchemaTypeInferenceException` for simpler error handling in future. 
RuntimeException is too abstract.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -196,6 +197,52 @@ public void doInsert(List<Record> records, QueryIdentifier queryId, ErrorReporte
...
+        if (csc.isAutoEvolve()) {
+            table = doInsertWithSchemaEvolution(records, table, queryId);
+        } else {
+            doInsertBatch(records, table, queryId);
+        }
+    }
+
+    private Table doInsertWithSchemaEvolution(List<Record> records, Table table, QueryIdentifier queryId) throws IOException, ExecutionException, InterruptedException {
+        // Split records into sub batches at schema boundaries (like JDBC BufferedRecords.add pattern)
+        // When schema changes mid batch the current sub batch is flushed, the table is evolved, and the insertion continues.
```
chernser: what problem does it solve?

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -196,6 +197,52 @@ public void doInsert(List<Record> records, QueryIdentifier queryId, ErrorReporte
         String database = first.getDatabase();
         Table table = getTable(database, topic);
         if (table == null) { return; }//We checked the error flag in getTable, so we don't need to check it again here
+
+        if (csc.isAutoEvolve()) {
+            table = doInsertWithSchemaEvolution(records, table, queryId);
```
chernser: after this we have two places making insert to maintain. In general we want to minimize number of methods we need to change. Now if `doInsertBatch` signature is change we have to update it in two places.
schema evolution just creates batches per schema update. So it can evolve schema and return batch to insert. It can be a loop. But `doInsertBatch` will be used in one place.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -196,6 +197,52 @@ public void doInsert(List<Record> records, QueryIdentifier queryId, ErrorReporte
...
+        }
+    }
+
+    private Table doInsertWithSchemaEvolution(List<Record> records, Table table, QueryIdentifier queryId) throws IOException, ExecutionException, InterruptedException {
+        // Split records into sub batches at schema boundaries (like JDBC BufferedRecords.add pattern)
+        // When schema changes mid batch the current sub batch is flushed, the table is evolved, and the insertion continues.
+        Schema currentSchema = getValueSchema(records.get(0));
+        int batchStart = 0;
+
+        for (int i = 1; i <= records.size(); i++) {
```
chernser: there is a loop over records at the start of the sink when we create data records. 
So we can detect need of schema change there. 
This is one of the reasons we are suspicious about benefits of schema evolution feature - to add a few columns we add more complexity.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -810,6 +860,88 @@ protected void doWriteCol(Data value, boolean fieldExists, Column col, OutputStr
...
+
+        for (String fieldName : missingColumns) {
+            Schema fieldSchema = schemaMap.get(fieldName);
+            if (fieldSchema == null) {
+                continue;
+            }
+
+            if (!fieldSchema.isOptional() && fieldSchema.defaultValue() == null) {
+                throw new RuntimeException(String.format(
+                        "Cannot auto-evolve: field '%s' is not optional and has no default value. " +
```
chernser: may be just create a nullable column?

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -810,6 +860,88 @@ protected void doWriteCol(Data value, boolean fieldExists, Column col, OutputStr
...
+                        "ClickHouse requires new columns to be either Nullable or have a DEFAULT.", fieldName));
+            }
+
+            String chType;
+            try {
+                chType = Column.connectTypeToClickHouseType(fieldSchema);
+            } catch (RuntimeException e) {
+                throw new RuntimeException(String.format(
```
chernser: As I've mentioned before - we can throw specific exception. Then no need to catch it here because it will point to code where it failed. 
besides this new exception doesn't add more information about exception.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -810,6 +860,88 @@ protected void doWriteCol(Data value, boolean fieldExists, Column col, OutputStr
...
+                        "Cannot auto-evolve: field '%s' has unsupported type for auto-evolution. %s", fieldName, e.getMessage()), e);
+            }
+
+            // ClickHouse does not allow Nullable wrapping for Array and Map types
+            if (fieldSchema.isOptional()
+                    && fieldSchema.type() != Schema.Type.ARRAY
+                    && fieldSchema.type() != Schema.Type.MAP) {
+                chType = "Nullable(" + chType + ")";
```
chernser: This is good! 
Then if we have mandatory field in connect schema we always have value and no need to limit to support of optional with default values.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -810,6 +860,88 @@ protected void doWriteCol(Data value, boolean fieldExists, Column col, OutputStr
...
+            }
+
+            columnDefs.add(String.format("`%s` %s", fieldName, chType));
```
chernser: Here we add just `fieldName Nullable(...)` and do not specify default. Am I right? 
It would be great if `connectTypeToClickHouseType` return complete column type without need to adjust it. Problem with constant adjusting is that we have to care about to many cases we may even do not know.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -810,6 +860,88 @@ protected void doWriteCol(Data value, boolean fieldExists, Column col, OutputStr
...
+        if (!columnDefs.isEmpty()) {
+            chc.alterTableAddColumns(table.getDatabase(), table.getCleanName(), columnDefs);
+            LOGGER.info("Schema evolution complete for table {}. Added columns: {}", table.getName(), columnDefs);
+            table = refreshTableAfterDDL(table, missingColumns);
+        }
+
+        return table;
+    }
+
+    private static final int DDL_REFRESH_MAX_RETRIES = 5;
```
chernser: should be configurable. 5 attempts is too many.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -810,6 +860,88 @@ protected void doWriteCol(Data value, boolean fieldExists, Column col, OutputStr
...
+            if (stillMissing.isEmpty()) {
+                return refreshed;
+            }
+            LOGGER.warn("DDL refresh attempt {}/{}: columns {} not yet visible, retrying in {}ms",
+                    attempt + 1, DDL_REFRESH_MAX_RETRIES, stillMissing, DDL_REFRESH_BACKOFF_MS);
+            try {
+                Thread.sleep(DDL_REFRESH_BACKOFF_MS);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException("Interrupted while waiting for DDL propagation", e);
```
chernser: it should be InterruptedException because other code may special handling for this.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -810,6 +860,88 @@ protected void doWriteCol(Data value, boolean fieldExists, Column col, OutputStr
...
+            }
+        }
+        // Final attempt, use whatever we have
+        LOGGER.error("DDL propagation timeout: some columns may not be visible yet after {} retries. Proceeding with latest table state.",
```
chernser: at this point we know it will fail. should we throw exception, may be `RetriableException`

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
guilleov: Fixed in commit https://github.com/ClickHouse/clickhouse-kafka-connect/pull/687/changes/259dbb812e795b83ce79febe459e58e900d117fe

src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Column.java
guilleov: Fixed https://github.com/guilleov/clickhouse-kafka-connect/commit/bb13864bd6dbf54f1819a33829d5c33d8654452a

src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Column.java
guilleov: Created here https://github.com/guilleov/clickhouse-kafka-connect/commit/319596142515174cea8da915514d5233652fa8f8

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: Modified in here https://github.com/guilleov/clickhouse-kafka-connect/commit/0638bdacd6c13505058aaed72412613510145245

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: Changed here https://github.com/guilleov/clickhouse-kafka-connect/commit/c8ff2eb8fdb257425acbe8b1a412ead83427c123

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: Changed here https://github.com/guilleov/clickhouse-kafka-connect/commit/c8ff2eb8fdb257425acbe8b1a412ead83427c123

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: Changed here https://github.com/guilleov/clickhouse-kafka-connect/commit/c8ff2eb8fdb257425acbe8b1a412ead83427c123

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: I was thinking it might fail if schema changed mid batch insert would fail because some records wouldn't have the correct schema but I think thats not the case right?

I did simplified this section on commit: https://github.com/guilleov/clickhouse-kafka-connect/commit/022f88e049610ee8f0d285ff44556716dcc38b55

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: Changed on commit: https://github.com/guilleov/clickhouse-kafka-connect/commit/022f88e049610ee8f0d285ff44556716dcc38b55

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: done: https://github.com/guilleov/clickhouse-kafka-connect/commit/15c995e34211d8f3f575bf2e2a53bd07b5c83b6f

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: Added throws InterruptedException https://github.com/guilleov/clickhouse-kafka-connect/commit/cdd622a4a0bc2c6dea212c4ae12f56ceccd95e3d

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: Changed to retriable https://github.com/guilleov/clickhouse-kafka-connect/commit/cdd622a4a0bc2c6dea212c4ae12f56ceccd95e3d

src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Column.java
guilleov: How we should handle unions?

guilleov: https://github.com/guilleov/clickhouse-kafka-connect/commit/a9586e54c3771a91ad6dca3ed7256a43221d9e6d dunno what you think of this way of handling it @chernser let me know

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -196,6 +199,18 @@ public void doInsert(List<Record> records, QueryIdentifier queryId, ErrorReporte
         String database = first.getDatabase();
         Table table = getTable(database, topic);
         if (table == null) { return; }//We checked the error flag in getTable, so we don't need to check it again here
+
+        if (csc.isAutoEvolve()) {
+            // New columns are Nullable, so older records without the new fields insert with NULL.
+            Record last = records.get(records.size() - 1);
```
mzitnik, line 219: how can we garante that the last recrod will have all fields?

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
```diff
@@ -1560,4 +1562,831 @@ public void testAvroDateAndTimeTypes() throws Exception {
             assertEquals(event.getTime2().atDate(LocalDate.of(1970, 1, 1)).format(localFormatter), row.get("time2"));
         }
     }
+
```
mzitnik, line 1808: Can we expand our tests to insert 3 batches:

Batch 1 → Schema V1
Batch 2 → Schema V2
Batch 3 → Schema V3

Also, add one batch with 10 records:

Records 1–5 with Schema V1 (3 fields)
Records 6–7 with Schema V2 (8 fields)
Records 8–10 with Schema V3 (5 fields)

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
guilleov: - If kafka topics are ordered last messages should have the newest schema. 
- If ignorePartitionsWhenBatching=true that might not be the case.
- If someones evolves the schema to V2 but then another producer keeps sending V1 messages since are backward compatible could potentially have a V1 -> V2 -> V1 schema insert and also fail to evolve

Might need to check all records and look for the new fields since we don't support deletion we could obtain a superset of all new fields. Imagine schema changes from V1 -> V2 -> V3 -> V4 in one batch we should keep new fields from V2/3/4 and generate the DDL based on that like if it was only one migration from V1 to a V4 with all fields

guilleov: Check https://github.com/guilleov/clickhouse-kafka-connect/commit/859e1626bfca039caad3aefeac0366d6567e89bc @mzitnik

mzitnik: This implementation is overkill to iterte all records again.
I would have a dislaimer in docs about this situation or implemented some other way

guilleov: Another way might be to let it fail when schema changes (schema should change often) and then retry and in that case we check all records to see what needs to be modified (Bad design to let a query fail just to retry I think but might work)

Other options are moving the detection of schema changes up to the ProxySinkTask.java where we iterate over all records already

What option do you think will be better for this new feature? @chernser @mzitnik

mzitnik: Yes, this is what I was thinking @guilleov, let's roll back to check only the last for now. 
Add this as a limitation, and we will come back to it later

guilleov: Ok, I reverted it and noted it both here and in the clickhouse-docs PR. Thanks @mzitnik 

https://github.com/guilleov/clickhouse-kafka-connect/commit/4ad5f17f749d1e354285c6831056b4b7bfca98cf

chernser (approved):

guilleov:
any updates? something else left to do @chernser @mzitnik @kurnoolsaketh

mzitnik (approved):

mzitnik:
@guilleov i have approved can just fix the conflicts

guilleov:
@mzitnik conflicts resolved!

mzitnik:
@guilleov is this passing for you localy? seems like tests are still failling

mzitnik:
@guilleov are sync with main there still errors in tests

guilleov:
let me check @mzitnik

guilleov:
@mzitnik sorry some test were outdated per my last change of checking only last record (and also some missing method that is not available on main)

Could you re-run the CI tests?

guilleov:
I did run locally now and worked

Tests summary: 265 tests, 262 succeeded, 0 failed, 3 skipped

chernser (approved):

Issue #689, opened 2026-03-25
[repo] Migrate to Gradle 9+


chernser:
### Description 

Today we are using gradle 7.+. For Gradle world it is ancient and we need to migrate at least to 9.3+ 

There are several reasons:
- dependabot cannot parse complicated versioning with variables. Gradle 9 has version catalog finally. 
- there are some outdated approaches that may stop working soon
- project structure might be groomed. For example integration tests.

PR #690, opened 2026-03-25
[security] Upgrade Jakson-core to 2.21.2 to address CVE  

chernser:
## Summary
There is a CVE in the version of `jakson-core` we currently use.  
Here is description https://github.com/advisories/GHSA-72hv-8253-57qq

This upgrade version of the package to a latest compatible one within same major version of `jakson` library. 
The library will be upgrade to `3.+` version later after testing. 

## Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -55,6 +55,7 @@ extra.apply {
     set("kafkaVersion", "2.7.0")
     set("gson", "2.13.1")
     set("jackson", "2.21.2")
+    set("jackson-annotations", "2.21")
 
     // Testing dependencies
     set("junitJupiterVersion", "5.9.2")
@@ -95,7 +96,7 @@ dependencies {
     // To parse JSON response from ClickHouse to parse complex data types correctly
     implementation("com.fasterxml.jackson.core:jackson-core:${project.extra["jackson"]}")
     implementation("com.fasterxml.jackson.core:jackson-databind:${project.extra["jackson"]}")
-    implementation("com.fasterxml.jackson.core:jackson-annotations:2.21}")
+    implementation("com.fasterxml.jackson.core:jackson-annotations:${project.extra["jackson-annotations"]}}")
 
     /*
         Will in side the Confluent Archive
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -95,7 +95,7 @@ dependencies {
     // To parse JSON response from ClickHouse to parse complex data types correctly
     implementation("com.fasterxml.jackson.core:jackson-core:${project.extra["jackson"]}")
     implementation("com.fasterxml.jackson.core:jackson-databind:${project.extra["jackson"]}")
-    implementation("com.fasterxml.jackson.core:jackson-annotations:${project.extra["jackson"]}")
+    implementation("com.fasterxml.jackson.core:jackson-annotations:2.21}")
```
kurnoolsaketh: why hardcode 2.21? Please revert this

build.gradle.kts
```diff
@@ -107,7 +107,7 @@ dependencies {
     clickhouseDependencies("com.google.code.gson:gson:${project.extra["gson"]}")
     clickhouseDependencies("com.fasterxml.jackson.core:jackson-core:${project.extra["jackson"]}")
     clickhouseDependencies("com.fasterxml.jackson.core:jackson-databind:${project.extra["jackson"]}")
-    clickhouseDependencies("com.fasterxml.jackson.core:jackson-annotations:${project.extra["jackson"]}")
+    clickhouseDependencies("com.fasterxml.jackson.core:jackson-annotations:2.21}")
```
kurnoolsaketh: same here

kurnoolsaketh (changes requested):

build.gradle.kts
chernser: because there is no `2.21` version for annotations. 
This kind of variables actually break dependabot because it cannot statically analyze dependencies. 
Here is issue to fix it for repo: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/689

kurnoolsaketh: you mean there is no `2.21.2` for annotations?

chernser: yes.

kurnoolsaketh: If we are changing to hardcoded versions in the dependency string, we should change them all - but this is out of scope for this PR. For now, I suggest that you add jackson-annotations to the `project.extra` map like all the other deps. Let's follow up with another PR to fix the dependency strings.

chernser: ok. fixed.

build.gradle.kts
chernser: fixed

Issue #672, opened 2026-03-11
[Integration Tests] Test against different Confluent platform versions


kurnoolsaketh:
Currently, all confluent docker images in ConfluentPlatform.java are pinned to version 7.5.0. We should test against multiple versions to spot and address changes in the confluent runtime that may break the connector.

chernser:
Blocked by https://github.com/ClickHouse/clickhouse-kafka-connect/issues/673

PR #697, opened 2026-03-27
Extend avro test coverage + identify compatible/incompatible schemas

### LINKED ISSUE #635: [test+doc] Extend testing with Avro messages
## Description 

We have got issues with unions. We need to extend tests for handling Avro originated records. 
Connector does not deal with Avro directly but such messages are converted to sink record and may have some specific implementation details 

Documentation should be updated. 

@chernser updated on April 3, 2026

This issue covers next things: 
- testing how all Avro types are converted to SinkRecord schema and how our Sink processes them 
- identify gaps and document them. Create another issue to address what is not implemented
- create a table how Avro types are converted to SinkRecord types and what types should be in destination columns. 

Note: there may be some limitations or things we need to discuss. In such case new issues should be created. This issue is only to understand current state of things.

kurnoolsaketh:
## Summary
In this PR:
- define an integration test harness to produce avro records to a kafka topic and sink them into ClickHouse
- add compatible and incompatible avro schemas. The incompatible schemas will be documented and investigated further.
    **- note: _this is meant to be an evolving set of schemas. As we extend the connector to support previously incompatible schemas, we should move them to `compatible` to ensure they are covered in the integration test runs._**

Closes: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/635

## Checklist

### DIFF
--- a/src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ConfluentPlatform.java
+++ b/src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ConfluentPlatform.java
@@ -1,35 +1,29 @@
 package com.clickhouse.kafka.connect.sink.helper;
 
 
-import okhttp3.MediaType;
-import okhttp3.OkHttpClient;
-import okhttp3.Request;
-import okhttp3.RequestBody;
-import okhttp3.Response;
+import okhttp3.*;
 import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
 import org.testcontainers.containers.wait.strategy.Wait;
-import org.testcontainers.shaded.com.fasterxml.jackson.databind.ObjectMapper;
 import org.testcontainers.utility.DockerImageName;
 import org.testcontainers.utility.MountableFile;
+
+import org.json.JSONArray;
 
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
 import java.util.regex.Pattern;
 
 
 public class ConfluentPlatform {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConfluentPlatform.class);
 
-    private static final String CONFLUENT_VERSION = "7.5.0";
+    private static final String CONFLUENT_VERSION = "7.5.0"; // TODO: this is >2 years outdated and should be upgraded or removed outright
     private static final DockerImageName KAFKA_REST_IMAGE = DockerImageName.parse(
             "confluentinc/cp-kafka-rest:" + CONFLUENT_VERSION
     );
@@ -67,6 +61,14 @@ public class ConfluentPlatform {
     private String restProxyEndpoint = null;
     private String connectRestEndPoint = null;
     private String ksqlRestEndPoint = null;
+
+    // ref: https://docs.confluent.io/platform/current/kafka-rest/api.html#post--topics-(string-topic_name)
+    // NOTE: the below keys are constants in Confluent rest proxy API v2.
+    private static final String AVRO_V2_APPLICATION_TYPE = "application/vnd.kafka.avro.v2+json";
+    private static final String TOPIC_ENDPOINT_RESPONSE_OFFSETS_KEY = "offsets";
+    private static final String TOPIC_ENDPOINT_RESPONSE_OFFSETS_ERROR_KEY = "error";
+    private static final String CONNECTORS_ENDPOINT_RESPONSE_TASKS_KEY = "tasks";
+    private static final String CONNECTORS_ENDPOINT_RESPONSE_TASKS_STATE_KEY = "state";
 
     GenericContainer<?> zookeeper;
     GenericContainer<?> cp_server;
@@ -263,8 +265,8 @@ public class ConfluentPlatform {
         Request request = new Request.Builder()
                 .url(clusterIDurl)
                 .build();
-        try (Response response = client.newCall(request).execute()) {
-            JSONObject obj = new JSONObject(response.body().string());
+        try (Response response = client.newCall(request).execute(); ResponseBody responseBody = response.body()) {
+            JSONObject obj = new JSONObject(responseBody.string());
             return obj.getJSONArray("data").getJSONObject(0).getString("cluster_id").toString();
         } catch (IOException ioe) {
             return null;
@@ -285,12 +287,9 @@ public class ConfluentPlatform {
                 .url(kafkaTopicEndpoint)
                 .post(body)
                 .build();
-        try (Response response = client.newCall(request).execute()) {
+        try (Response response = client.newCall(request).execute(); ResponseBody responseBody = response.body()) {
             LOGGER.debug("Create topic response code: {}", response.code());
-            assert response.body() != null;
-            String responseBody = response.body().string();
-            LOGGER.debug("Create topic response body: {}", responseBody);
-            return responseBody;
+            return responseBody.string();
         }
     }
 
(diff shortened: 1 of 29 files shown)


src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
```diff
@@ -1802,4 +1803,420 @@ public void testAvroDateAndTimeTypes() throws Exception {
             assertEquals(event.getTime2().atDate(LocalDate.of(1970, 1, 1)).format(localFormatter), row.get("time2"));
         }
     }
+
+    @Test
+    public void testAvroAllPrimitiveTypes() throws Exception {
+        final String topic = createTopicName("test_avro_primitives");
+
+        org.apache.avro.Schema avroSchema = new org.apache.avro.Schema.Parser().parse(
+                "{\"type\":\"record\",\"name\":\"PrimitiveRecord\",\"namespace\":\"com.clickhouse.test\"," +
```
chernser: this is very hard to maintain. 
Previously I had a avro messages in resource and compiled classes. Look into textFixtures https://github.com/ClickHouse/clickhouse-kafka-connect/tree/main/src/testFixtures/java/com/clickhouse/kafka/connect/avro/test

Ideally I would like to see parameterized tests where input is avro message with schema and output is sink records that are fed to Sink Task and rows written in DB. 
1. we need to use real converters from kafka. So it might be an integration tests. We will test with different version later. 
2. we need to get result from converter to understand and debug SinkRecord. This is needed for issue fixing.
3. Finally we need to verify that connector can handle (easy part :-))

chernser: saving to db should be also decoupled - we may want to try different destination datatypes

chernser (changes requested):

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
kurnoolsaketh: > This is needed for issue fixing.

@chernser which issues? any specific examples would be helpful, as I generated these test cases using claude.

chernser: In the past I had two issues about unions in avro and protobuf. 
I've created a real avro record and converted with actual converter. 
Intention was to have readable code. 

Now imagine adding more fields to string that represents schema in this PR. Then I need someone to review. 
Yes, we now have AI to do this work but: 
- we still have to review code and less code - less cognitive load
- we still have to know what is tested so need to keep inventory compact. 
- tests sometimes act as reference code.

kurnoolsaketh: > we need to use real converters from kafka.

@chernser So you specifically want to test kafka connect's converter from avro -> SinkRecord? If so, why should this be in clickhouse connector tests? This feels more like a unit test of kafka connect.

Basically, i'm trying to understand why the connector has issues with avro specifically - is there something special about how avro is converted to SinkRecord that breaks the connector?

kurnoolsaketh: > we need to use real converters from kafka.

@chernser also re this, kafka connect uses `AvroConverter.java` under the hood, which is what we use in the unit test `ClickHouseSinkTaskWithSchemaTest.java`. To me this means if avro unions work in our unit tests, they should work in a running connector, unless avro serialization is somehow kafka connect version specific. So, is `testAvroWithUnion` https://github.com/ClickHouse/clickhouse-kafka-connect/blob/d1d4839021af93dadff69e2a15ce17baaab494ba/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java#L1707 not sufficient?

chernser: I want to test our connector with actual output of kafka converter. This is critical because we need to understand exactly what is passed to the Sink. 
We need to make a modular chain to be able to adjust different parts of it like source message, converter version, may be some environment. 

We do not test converter we use it to create test data because it is closest way to production.

mzitnik:
@kurnoolsaketh, what is the status of it? It seems stuck i see @chernser asked for changes two weeks ago

kurnoolsaketh:
@mzitnik still in progress, marking it as draft for now

kurnoolsaketh:
it run: https://github.com/ClickHouse/clickhouse-kafka-connect/actions/runs/24427974178

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ConfluentPlatform.java
```diff
@@ -1,35 +1,29 @@
...
-import java.util.UUID;
+import java.util.*;
 import java.util.regex.Pattern;
 
 
 public class ConfluentPlatform {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConfluentPlatform.class);
 
-    private static final String CONFLUENT_VERSION = "7.5.0";
+    private static final String CONFLUENT_VERSION = "7.5.0"; // TODO: this is >2 years outdated and should be upgraded or removed outright
```
chernser, line 26: support should be properly deprecated. We may not drop support even for very old things. 
https://docs.confluent.io/platform/current/installation/versions-interoperability.html#cp-and-apache-ak-compatibility

chernser (approved):

chernser:
@kurnoolsaketh 
Thank you for the PR! 
Would you please add some documentation of how it works? 

Thanks!

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ConfluentPlatform.java
kurnoolsaketh: @chernser do you have any objection to just upgrading confluent version 7.9.6 now? This is a patch version that's valid until at least 2027.

kurnoolsaketh:
@chernser i added a comment explaining the test and how to add to it

kurnoolsaketh:
tracked fixing incompatible schemas here: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/726

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ConfluentPlatform.java
chernser: @kurnoolsaketh 
In tests we should do it. Like with ClickHouse we need to test different versions. 
But we should not update other dependencies until we understand how it affects users.

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ConfluentPlatform.java
```diff
@@ -263,8 +265,8 @@ private String extractClusterId() {
         Request request = new Request.Builder()
                 .url(clusterIDurl)
                 .build();
-        try (Response response = client.newCall(request).execute()) {
-            JSONObject obj = new JSONObject(response.body().string());
+        try (Response response = client.newCall(request).execute(); ResponseBody responseBody = response.body()) {
+            JSONObject obj = new JSONObject(responseBody.string());
```
mzitnik, line 269: If responseBody == null, we will not catch it correct me if i'am wrong

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ConfluentPlatform.java
kurnoolsaketh: test version is upgraded now 👍

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ConfluentPlatform.java
kurnoolsaketh: ResponseBody is non-null (pasted the javadoc comment below):

> Returns a non-null stream with the server's response. The returned value must be closed and may be consumed only once.

mzitnik (approved):
Approving, maybe we can open an issue if we do not cover it. current the implementation is for Confluent, a specific version. I think we should make it configurable

kurnoolsaketh:
@mzitnik tracked here: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/672

Issue #643, opened 2025-11-24
CI - Run tests with ClickHouse cluster


BentsiLeviav:
### Description

The test setup is designed for integration testing of the Connector, focusing on single-instance ClickHouse testing rather than a clustered environment. Running the tests regularly with a cluster setup will help us identify mismatches early on during the development phase.

PR #705, opened 2026-04-03
consolidate test helpers into `ClickHouseTestHelpers`

### LINKED ISSUE #704: Standardize test helper functions in `ClickHouseTestHelpers.java`
Currently, test helpers that create/alter/drop/optimize tables, count rows, and etc are duplicated and spread throughout the test code in a few places:
- ClickHouseBase
- ClickHouseTestHelpers
- ClickHouseAPI
- test implementations themselves

this makes it difficult, repetitive, and error prone to add additional functionality to them (such as injecting DDL, settings, and extending their usage for other tests). Fix this by moving and standardizing all helper functions in ClickHouseTestHelpers. This will make developing tests much easier, clearer, and less error prone.

kurnoolsaketh:
# https://github.com/ClickHouse/clickhouse-kafka-connect/issues/643 PR 1/3
## Summary
Currently, test helpers that create/alter/drop/optimize tables, count rows, and etc are duplicated, leak HTTP connections in some cases, and spread throughout the test code in a few places:
- ClickHouseBase
- ClickHouseTestHelpers
- ClickHouseAPI
- test implementations themselves

This makes it difficult, repetitive, and error prone to add additional functionality to them (such as injecting DDL, settings, and extending their usage for other tests) and to bug-fix them in one place. Fix this by moving and standardizing all helper functions in ClickHouseTestHelpers. This will make developing tests much easier, clearer, and less error prone.

Also, in preparation for the cluster testing PRs, rename current CI tests to reflect `standalone` mode (ie, tests run against a single container CH db).

Closes #704 

## Checklist

### DIFF
--- a/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
+++ b/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
@@ -54,16 +54,7 @@ public class ClickHouseBase {
 
     @AfterAll
     protected void tearDown() {
-        // TODO: disable dropping database for debug
-        if (isCloud) { // We need to clean up databases in the cloud, we can ignore the local database
-            if (!ClickHouseTestHelpers.DATABASE_DEFAULT.equals(database)) {
-                try (var tmpClient = ClickHouseTestHelpers.createClient(getBaseProps())) {
-                    ClickHouseTestHelpers.dropDatabase(tmpClient, database);
-                } catch (Exception e) {
-                    LOGGER.error("Error dropping database", e);
-                }
-            }
-        } else {
+        if (!isCloud) {
             ClickHouseContainer ch = getDb();
             if (ch != null) {
                 LOGGER.info("Stopping db container: id={}, port={}", ch.getContainerId(), ch.getMappedPort(8123));


kurnoolsaketh:
IT run: https://github.com/ClickHouse/clickhouse-kafka-connect/actions/runs/23966570467

.github/workflows/tests.yaml
```diff
@@ -16,20 +16,50 @@ on:
...
       matrix:
         client: ["V1", "V2"]
-        clickhouse: ["25.3", "25.8", "25.10", "25.11", "25.12", "latest", "cloud"]
-    name: ClickHouse ${{ matrix.clickhouse }} Client version ${{ matrix.client }} tests
+        clickhouse: ["25.3", "25.8", "25.10", "25.11", "25.12", "latest"]
+    name: ClickHouse Standalone ${{ matrix.clickhouse }} Client version ${{ matrix.client }} tests
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
```
chernser: please upgrade to v4 or v5 
here is official documentation https://docs.github.com/en/actions/tutorials/build-and-test-code/java-with-maven

.github/workflows/tests.yaml
```diff
@@ -16,20 +16,50 @@ on:
...
+      - name: Run Tests
+        env:
+          CLICKHOUSE_VERSION: ${{ matrix.clickhouse }}
+          CLIENT_VERSION: ${{ matrix.client }}
+        run: ./gradlew test
+      - name: Upload test results
+        uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: result standalone_${{ matrix.client }}_${{ matrix.clickhouse }}
```
chernser: as we can run several in parallel - artifact should have a job id

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
chernser: optional: would you please remove this duplication with building string and making auth header?

chernser: optional: we could use V2 in all tests I think.

chernser: optional: there should be only time limit. error message is very useless because in the logs it will be unclear where we failed. need to be something "Failed to restart service in time"

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
```diff
@@ -263,7 +264,7 @@ public Records queryV2(String query) {
         return queryV2(query, (ClickHouseFormat) null);
     }
 
-    public Records queryV2(String query, @Nullable ClickHouseFormat clickHouseFormat) {
+    public Records queryV2(String query, ClickHouseFormat clickHouseFormat) {
```
chernser: optional: may be document what format will be used if no specified? 
Actually I think it will lead us to some problem because not all formats are supported by clients. So here we may be need to hardcode format and say what is it (currently code I think passes only one).

src/test/java/com/clickhouse/kafka/connect/sink/dlq/FailureTest.java
```diff
@@ -37,8 +37,8 @@ public class FailureTest extends ClickHouseBase {
     }
     @Test
     void testSchemaValidationFailure() throws Exception {
-        Map<String, String> props = createProps();
-        ClickHouseHelperClient chc = createClient(props);
+        Map<String, String> props = getBaseProps();
```
chernser, line 40: optional: if getBaseProps() accept map with custom properties we could join it with createClient and replace this lines with single method call. Result would be small class with actual props and client helper.

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
```diff
@@ -27,61 +22,57 @@ public class ClickHouseBase {
...
 
     @AfterAll
     protected void tearDown() {
         // disable dropping database for debug
-        if (isCloud) {//We need to clean up databases in the cloud, we can ignore the local database
-            if (database != null) {
-                try {
-                    dropDatabase(getDatabase());
+        if (isCloud) { // We need to clean up databases in the cloud, we can ignore the local database
+            if (database != null && !database.equals(ClickHouseTestHelpers.DATABASE_DEFAULT)) {
```
chernser: I'd put some warning here.

actually it should be checked on the start that database is not default.

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaProxyTest.java
```diff
@@ -384,7 +384,7 @@ public void schemaWithDecimalTest() {
     @Test
     public void schemaWithBytesTest() {
         Map<String, String> props = getTestProperties();
-        ClickHouseHelperClient chc = createClient(props, true);
```
chernser, line 387: this is intentional because there are two options to set db - via operation settings and via client config. 
This `true` tells that we want to test how it works if default db is set in config. If needed this changes - then should be in separate PR.

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
```diff
@@ -26,53 +28,61 @@
...
     }
 
     public static boolean isCloud() {
-        String version = System.getenv("CLICKHOUSE_VERSION");
+        String version = System.getenv(CLICKHOUSE_VERSION);
         LOGGER.info("Version: {}", version);
         return version != null && version.equalsIgnoreCase("cloud");
     }
 
     public static void query(ClickHouseHelperClient chc, String query) {
```
chernser: optional: very very strange method - query without result set. :-) 
May be return if empty or so. May be rename to `executeUpdate` ?

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
```diff
@@ -26,53 +28,61 @@
...
-            } catch (Exception e) {
-                LOGGER.info("Failed to query ", e);
-                throw new RuntimeException(e);
-            }
-        } else {
-            chc.queryV1(query);
+        try (Records ignored = chc.queryV2(query)) {
+            // success
+        } catch (Exception e) {
+            LOGGER.info("Failed to query ", e);
```
chernser: lest add query as it is a test code.

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
```diff
@@ -201,6 +211,18 @@ public static OperationMetrics optimizeTable(ClickHouseHelperClient chc, String
         }
     }
 
+    public static int countRows(ClickHouseHelperClient chc, String database, String topic) {
+        String queryCount = String.format("select count(*) from `%s.%s`", database, topic);
+        try (ClickHouseClient client = ClickHouseClient.newInstance(ClickHouseProtocol.HTTP);
```
chernser: can we use v2 client here?  
`queryAll()` would be shortest way.

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
```diff
@@ -395,4 +417,125 @@ public static void runQuery(ClickHouseHelperClient chc, String query) {
...
+    public static void doPing(ClickHouseHelperClient chc) {
+        boolean ping;
+        int retry = 0;
+
+        do {
+            LOGGER.info("Pinging ClickHouse server to warm up for tests...");
+            ping = chc.ping();
+            if (!ping) {
+                LOGGER.info("Unable to ping ClickHouse server, retrying... {}", retry);
+            }
```
chernser: ping has retry already. So this code multiplies attempts.

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
```diff
@@ -395,4 +417,125 @@ public static void runQuery(ClickHouseHelperClient chc, String query) {
...
+            throw new RuntimeException("Unable to ping ClickHouse server...");
+        }
+    }
+
+    public static ClickHouseHelperClient createClient(Map<String, String> props) {
+        ClickHouseSinkConfig csc = new ClickHouseSinkConfig(props);
+
+        String hostname = csc.getHostname();
+        int port = csc.getPort();
+        String database = csc.getDatabase();
```
chernser: this seems unused and differs from the original method. Please revert - we need to review those cases when database should be set.

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
```diff
@@ -395,4 +417,125 @@ public static void runQuery(ClickHouseHelperClient chc, String query) {
...
+        String username = csc.getUsername();
+        String password = csc.getPassword();
+        boolean sslEnabled = csc.isSslEnabled();
+        int timeout = csc.getTimeout();
+        String clientVersion = csc.getClientVersion();
+
+        return new ClickHouseHelperClient.ClickHouseClientBuilder(hostname, port, csc.getProxyType(), csc.getProxyHost(), csc.getProxyPort())
+                .setDatabase(csc.getDatabase())
```
chernser, line 367: previous method set `default` here.

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
```diff
@@ -395,4 +417,125 @@ public static void runQuery(ClickHouseHelperClient chc, String query) {
...
+
+        if (!ping) {
+            throw new RuntimeException("Unable to ping ClickHouse server...");
+        }
+    }
+
+    public static ClickHouseHelperClient createClient(Map<String, String> props) {
+        ClickHouseSinkConfig csc = new ClickHouseSinkConfig(props);
+
+        String hostname = csc.getHostname();
```
chernser: why do we need all these variables? please use inline.

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
kurnoolsaketh: what does v2 mean here? maybe you are confusing this client (java.net.http.HttpClient) with clickhouse client v2?

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
kurnoolsaketh: > actually it should be checked on the start that database is not default.

database is set to `kafka_connect_test_%d_%s` before each test class runs. "default" is never used in ClickHouseBase tests. The purpose of instantating database to "default" is to prevent null pointers in case any test tries to ping the server using client v1 but a database is not configured on the client yet.

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
chernser: this particular code uses V1 client. 
I would like to use only V2 in tests

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
chernser: may be set non existing database to catch case when tests misconfigured? 

There are three tests that I've mentioned in [src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaProxyTest.java](https://github.com/ClickHouse/clickhouse-kafka-connect/pull/705/files/9b6633e94696edfbe3726a679915f37bbc49b47e#diff-e98aaf60ef9b7fbcd71b84ce849b414f0d01d0b4871c3c39a941f2979da456a5) that require setting DB - by default we do not set it.

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaProxyTest.java
kurnoolsaketh: > This true tells that we want to test how it works if default db is set in config. If needed this changes - then should be in separate PR.

I'm not following you here. This is not changing the behavior of the tests because after this PR, database creation will happen explicitly before all tests - see: https://github.com/ClickHouse/clickhouse-kafka-connect/pull/705/changes#diff-f223978a7cb348e0bbf10b449b461d3bfcbd817310a2ccb998a7b35d2484e9ceR35

I removed this boolean argument to force users to create database explicitly instead of as a side-effect of client creation (the old way). In the old way, it's not transparent which database is being [created](https://github.com/ClickHouse/clickhouse-kafka-connect/blob/d1d4839021af93dadff69e2a15ce17baaab494ba/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java#L141) and it's redundant to `createDatabase` in every test that the client is created.

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
kurnoolsaketh: i will use queryV2 throughout this helper file to make the implementation uniform

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
kurnoolsaketh: ok, i've removed doPing since it's redundant

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
kurnoolsaketh: it is not unused, and the behavior is the same as the original because we are explicitly creating the database. See: https://github.com/ClickHouse/clickhouse-kafka-connect/pull/705#discussion_r3041551603

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
kurnoolsaketh: done

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
kurnoolsaketh: this file does not use clickhouse client at all anymore, so we are good here 👍

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
kurnoolsaketh: changed

.github/workflows/tests.yaml
kurnoolsaketh: done

.github/workflows/tests.yaml
kurnoolsaketh: done

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
kurnoolsaketh: done

kurnoolsaketh: done

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
kurnoolsaketh: > by default we do not set it.

are you sure? [`withDatabase` is set to true by default](https://github.com/ClickHouse/clickhouse-kafka-connect/blob/d1d4839021af93dadff69e2a15ce17baaab494ba/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java#L114) - all tests in ClickHouseSinkTaskWithSchemaProxyTest.java use `withDatabase=true` [except for `proxyPingTest`](https://github.com/ClickHouse/clickhouse-kafka-connect/blob/d1d4839021af93dadff69e2a15ce17baaab494ba/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaProxyTest.java#L125)

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private HttpResponse<String> executeRequest(String url, String method, HttpRequest.BodyPublisher body) throws URISyntaxException, IOException, InterruptedException {
+        String basicAuthCreds = Base64.getEncoder().encodeToString((properties.getProperty(CLICKHOUSE_CLOUD_API_KEY) + ":" + properties.getProperty(CLICKHOUSE_CLOUD_API_SECRET)).getBytes());
+        HttpRequest.Builder requestBuilder = HttpRequest.newBuilder()
+                .uri(new URI(url))
+                .header("Content-Type", "application/json;charset=UTF-8")
+                .header("Authorization", "Basic " + basicAuthCreds);
+
+        switch (method) {
+            case "PATCH": {
```
chernser, line 101: hmm... `PATCH` method ... now I understand why V1 is used :-) should we add it to V2?

src/integrationTest/java/com/clickhouse/kafka/connect/sink/ExactlyOnceTest.java
```diff
@@ -185,10 +187,29 @@ private void checkSpottyNetworkSchemaless(String topicName, int numberOfPartitio
...
-            ClickHouseAPI.dropTable(chcNoProxy, topicName);
+            ClickHouseTestHelpers.dropTable(chcNoProxy, topicName);
             runCount++;
         } while (runCount < 3 && allSuccess);
 
         assertTrue(allSuccess);
     }
+
+    private static Records selectDuplicates(ClickHouseHelperClient chc, String tableName) {
+        String queryString = String.format("SELECT `side`, `quantity`, `symbol`, `price`, `account`, `userid`, `insertTime`, COUNT(*) " +
```
chernser: format + builder - that is why I do not understand `String.format` in Java :-)

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
```diff
@@ -294,34 +284,15 @@ public static boolean validateRows(ClickHouseHelperClient chc, String topic, Col
...
-                System.out.println(r.getValue(0));
-            });
-        } catch (ClickHouseException e) {
-            throw new RuntimeException(e);
+    public static int countInsertQueries(ClickHouseHelperClient chc, String topic) throws Exception {
+        String sql = String.format("SELECT COUNT(*) " +
+                "FROM system.query_log " +
+                "WHERE type = 'QueryFinish' " +
+                "AND query_kind = 'Insert' " +
+                "AND executeQueryIgnoreResult ILIKE '%%%s%%'", topic);
```
chernser, line 293: hm...

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
chernser: is it done in code or new issue created to cover optional comments?

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
chernser: actually this method can be private because used only in class.

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
chernser: here is more:
com.clickhouse.kafka.connect.sink.ClickHouseSinkTaskWithSchemaProxyTest#withLowCardinalityTest
com.clickhouse.kafka.connect.sink.ClickHouseSinkTaskWithSchemaProxyTest#withUUIDTest

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaProxyTest.java
chernser: ok.

.github/workflows/tests.yaml
```diff
@@ -16,20 +16,50 @@ on:
       - 'docs/**'
 
 jobs:
-  build:
+  standalone-tests:
```
mzitnik, line 19: Why have we split it into two parts? ClickHouse version and cloud?

src/testFixtures/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseTestHelpers.java
kurnoolsaketh: see: https://github.com/ClickHouse/clickhouse-kafka-connect/pull/705#discussion_r3041551603

in the previous implementation, `default` was overridden by the database name, if one was set. Now, it's more transparent: whichever database is configured in ClickHouseSinkConfig will be the database in the client object.

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
kurnoolsaketh: what do you mean? this is the endpoint the test is querying: https://clickhouse.com/docs/cloud/manage/api/swagger#tag/Service/operation/instanceStateUpdate

src/integrationTest/java/com/clickhouse/kafka/connect/sink/ExactlyOnceTest.java
kurnoolsaketh: yeah this is terrible 🤣 . fixed

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
kurnoolsaketh: should we even keep this method? It's unused in the entire project except for the overloaded `queryV2(String query)` and kind of redundant since callers can use `settings.setFormat` themselves.

chernser: yes we might remove it if so.

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
kurnoolsaketh: i'm not sure what you are referring to here. currently, `createClient(props, true)` is the exact same as `createClient(props)`. There are only two instances of `createClient(props, false)`:

- ClickHouseSinkTaskSchemalessProxyTest.proxyPingTest
- ClickHouseSinkTaskWithSchemaProxyTest.proxyPingTest

these don't require database creation because they are ping tests. After this PR, the behavior of `ClickHouseTestHelpers.createClient` (new) will be the same as `ClickHouseBase.createClient(props, false)`. Users need to explicitly create the database by calling `ClickHouseTestHelpers.createDatabase()`. Are we on the same page? please lmk if there is something I'm missing.

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
kurnoolsaketh: done

src/test/java/com/clickhouse/kafka/connect/sink/dlq/FailureTest.java
kurnoolsaketh: I'm hesitant to do this because createClient is now extracted to the generic ClickHouseTestHelpers and getBaseProps() is specific to the unit tests (ClickHouseBase).

kurnoolsaketh: In some tests, the props are not even passed to createClient, so i think it's best to keep them separate.

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
chernser: Mainly interested if we need to set method type in client (patch is not much different from POST or PUT).

kurnoolsaketh: yeah according to the openapi spec we need to set it

kurnoolsaketh:
reason for failing CI: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/711

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public HttpResponse<String> stopInstance(String serviceId) throws URISyntaxException, IOException, InterruptedException {
+        return updateServiceState(serviceId, "stop");
+    }
+
+    public HttpResponse<String> startInstance(String serviceId) throws URISyntaxException, IOException, InterruptedException {
+        return updateServiceState(serviceId, "start");
+    }
+
+    public HttpResponse<String> updateServiceState(String serviceId, String command) throws URISyntaxException, IOException, InterruptedException {
```
mzitnik: lets have the command as Enum so it would be more strict

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
```diff
@@ -0,0 +1,117 @@
...
+        String serviceId = properties.getProperty(CLICKHOUSE_CLOUD_SERVICE_ID);
+        //1. Stop Instance
+        stopInstance(serviceId);
+
+        //2. Wait
+        String serviceState = getServiceState(serviceId);
+        int loopCount = 0;
+        while(!serviceState.equals("STOPPED") && loopCount < 60) {
+            LOGGER.debug("Service State: {}", serviceState);
+            Thread.sleep(5 * 1000);
```
mzitnik, line 69: lets extarct to const

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
```diff
@@ -0,0 +1,117 @@
...
+                break;
+            }
+            default: throw new RuntimeException("Unsupported request method: " + method);
+        }
+
+        return HttpClient.newBuilder().proxy(ProxySelector.getDefault()).build().send(requestBuilder.build(), HttpResponse.BodyHandlers.ofString());
+    }
+
+    private String getBaseServiceEndpoint(String serviceId) {
+        return "https://" + properties.getProperty(CLICKHOUSE_CLOUD_API_HOST) + "/v1/organizations/" + properties.getProperty(CLICKHOUSE_CLOUD_ORGANIZATION) + "/services/" + serviceId;
```
mzitnik: `String.format` is clearer here than string concatenation

src/integrationTest/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConnectorIntegrationTest.java
```diff
@@ -60,10 +56,10 @@ public static void setup() {
         connectorPath.add(confluentArchive);
         confluentPlatform = new ConfluentPlatform(network, connectorPath);
 
-        db = new ClickHouseContainer(ClickHouseTestHelpers.CLICKHOUSE_DOCKER_IMAGE).withNetwork(network).withNetworkAliases(CLICKHOUSE_DB_NETWORK_ALIAS);
+        db = new ClickHouseContainer(ClickHouseTestHelpers.CLICKHOUSE_DOCKER_IMAGE).withNetwork(network).withNetworkAliases(ClickHouseTestHelpers.CLICKHOUSE_DB_NETWORK_ALIAS);
```
mzitnik, line 59: why new line

src/integrationTest/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConnectorIntegrationTest.java
```diff
@@ -60,10 +56,10 @@ public static void setup() {
...
         db.start();
 
-        toxiproxy = new ToxiproxyContainer(TOXIPROXY_DOCKER_IMAGE_NAME).withNetwork(network).withNetworkAliases(TOXIPROXY_NETWORK_ALIAS);
+        toxiproxy = new ToxiproxyContainer(ClickHouseTestHelpers.TOXIPROXY_DOCKER_IMAGE_NAME).withNetwork(network).withNetworkAliases(ClickHouseTestHelpers.TOXIPROXY_NETWORK_ALIAS);
```
mzitnik, line 62: Why new line

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
```diff
@@ -27,61 +22,57 @@ public class ClickHouseBase {
...
-        // disable dropping database for debug
-        if (isCloud) {//We need to clean up databases in the cloud, we can ignore the local database
-            if (database != null) {
-                try {
-                    dropDatabase(getDatabase());
+        // TODO: disable dropping database for debug
+        if (isCloud) { // We need to clean up databases in the cloud, we can ignore the local database
+            if (!ClickHouseTestHelpers.DATABASE_DEFAULT.equals(database)) {
+                try (var tmpClient = ClickHouseTestHelpers.createClient(getBaseProps())) {
+                    ClickHouseTestHelpers.dropDatabase(tmpClient, database);
```
mzitnik: Why we want to drop the database in the cloud for our cloud instance we have a cleanup job 
It can be usafull for debug in case we have some error

.github/workflows/tests.yaml
kurnoolsaketh: @mzitnik last line in the PR description: `Also, in preparation for the cluster testing PRs, rename current CI tests to reflect standalone mode (ie, tests run against a single container CH db).`

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
kurnoolsaketh: done

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
kurnoolsaketh: done

src/integrationTest/java/com/clickhouse/kafka/connect/sink/helper/ClickHouseCloudAPI.java
kurnoolsaketh: done

src/integrationTest/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConnectorIntegrationTest.java
kurnoolsaketh: there's no new line here

src/integrationTest/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConnectorIntegrationTest.java
kurnoolsaketh: same here

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
kurnoolsaketh: this is a slight inconsistency in our test setup, but i think for a good reason. This logic only applies to CI tests that run against cloud - given that these tests run for every PR, I am hesitant to keep these databases around indefinitely. These tests are also straightforward to run against cloud from a local environment.

on the other hand, integration tests create all tables in default database and tables are not dropped if the test fails.

.github/workflows/tests.yaml
kurnoolsaketh: i will add a separate part for cluster tests in the follow-up PR for clarity and ease of debugging

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseBase.java
BentsiLeviav: > I am hesitant to keep these databases around indefinitely

@kurnoolsaketh - Given @mzitnik comment: "we have a cleanup job" - aka - We have a cleanup job that cleans all databases and tables generated by tests - what is the rationale behind your hesitation about indefinitely databases?

kurnoolsaketh: I see, I misunderstood @mzitnik's comment. Sure, that makes sense to me - I'll keep the databases and rely on the job to clean them async. 👍

kurnoolsaketh: done

PR #703, opened 2026-04-01
Agents md

mshustov:
## Summary
- make https://github.com/ClickHouse/clickhouse-kafka-connect/tree/main/docs AI friendly
- create `agents.md` based on `docs/` and https://clickhouse.com/blog/kafka-connect-connector-clickhouse-with-exactly-once

PR #710, opened 2026-04-07
Add claude-reviewer GH action

kurnoolsaketh:
## Summary
- This workflow gives claude clear permissions, objectives, and rules for the agent in how it should respond to users in a PR. This way, we have tight control over the prompt and the expected outputs. 💪
- It also includes a system prompt to increase the likelihood of certain actions by default (for ex, Do NOT change the PR title, description, or automatically push commits to this branch without explicitly asking the commenter beforehand).
- It prevents abuse by checking the commenter's permissions and asserts that only writers/admins may invoke claude reviews.
- It invokes the /review-pr command, which launches specialized, parallel subagents to handle different aspects of the code review. See: https://github.com/anthropics/claude-code-action/blob/6cad158a175744eb2e76f7f5fd108ec63145598c/.claude/commands/review-pr.md?plain=1#L3

## Example usage in a PR comment
- "@claude" -> should review the PR
- "@claude review this PR"
- "@claude where is the best place to move this helper function to prevent code duplication?"

### DIFF
--- a/.github/workflows/claude-reviewer.yml
+++ b/.github/workflows/claude-reviewer.yml
@@ -1,0 +1,86 @@
+name: claude-review-assistant.yml
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  issues:
+    types: [opened, assigned]
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  claude:
+    if: |
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
+      (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    env:
+      GH_TOKEN: ${{ github.token }}
+      # PR number lives in different places depending on the event type
+      PR_NUMBER: ${{ github.event.issue.number || github.event.pull_request.number }}
+      COMMENTER: ${{ github.event.comment.user.login }}
+      REPO: ${{ github.repository }}
+      # Reactions API endpoint differs: issue comments vs PR review comments
+      COMMENT_REACTIONS_URL: ${{ github.event_name == 'pull_request_review_comment' && format('repos/{0}/pulls/comments/{1}/reactions', github.repository, github.event.comment.id) || format('repos/{0}/issues/comments/{1}/reactions', github.repository, github.event.comment.id) }}
+
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+      id-token: write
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-reviewer.yml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          #          ref:
+          #          - https://github.com/anthropics/claude-code-action/blob/main/docs/solutions.md#automatic-pr-code-review
+          #          - https://github.com/anthropics/claude-code-action/blob/6cad158a175744eb2e76f7f5fd108ec63145598c/.claude/commands/review-pr.md?plain=1#L4
+          prompt: |
+            REPO: ${{ github.repository }}
+            PR_NUMBER: ${{ github.event.pull_request.number }}
+            
+            OBJECTIVE:
+            Read the comment you were tagged in and respond to the user. Gather relevant context for your answers from:
```
chernser, line 66: I would not say "respond" here because we need more certain actions like follows "gather relevant context". 
I'm not a big expert in this but instructions seems very broad. I'd say - "analyze user input and if it is about review start it according to AGENTS.md" 
For simplicity we should have one entry point for instructions. and also we need to limit what file it is because PR can add some.

.github/workflows/claude-reviewer.yml
```diff
@@ -0,0 +1,86 @@
...
+            RULES:
+            - Use `gh pr comment` for top-level feedback.
+            - Use `mcp__github_inline_comment__create_inline_comment` (with `confirmed: true`) to highlight specific code issues.
+            - Only post GitHub comments - don't submit review text as messages.
+            
+            Note: The PR branch is already checked out in the current working directory.
+
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          claude_args: |
+            --max-turns 10
```
chernser, line 83: lets start with 3 - 5

.github/workflows/claude-reviewer.yml
```diff
@@ -0,0 +1,86 @@
...
+            --model claude-opus-4-6
+            --system-prompt "You are an agent assisting users in reviewing PR's in this repository. Follow the coding patterns and standards of the project. Ensure all new code you propose has tests. Do NOT change the PR title, description, or automatically push commits to this branch without explicitly asking the commenter beforehand."
```
chernser, line 85: how it combines with main prompt? should we reference AGENTS.md here?

mshustov:
@kurnoolsaketh, should we prioritize merging https://github.com/ClickHouse/clickhouse-kafka-connect/pull/703 first to immediately test the setup?

kurnoolsaketh:
> @kurnoolsaketh, should we prioritize merging https://github.com/ClickHouse/clickhouse-kafka-connect/pull/703 first to immediately test the setup?


@mshustov yeah let's do that - I have some bigger changes I want to propose to the AGENTS file, but i can follow up with that in another PR.

BentsiLeviav:
@kurnoolsaketh As we discussed last week, this is pending internal discussion to figure out things like:
* Which model do we want to use? Why that over the others? 
* What type of tasks would be done by these agents? (CR, code analysis, auto tagging, size estimations, risk asssetment and more)
* Workflows reuse

Closing this for now, we will come back to that in the near future.

PR #717, opened 2026-04-10
Add local cluster testing infrastructure

### LINKED ISSUE #715: Add local cluster testing infrastructure
Add docker compose cluster setup (3-node ClickHouse with Keeper) and associated code components to enable testing against local clickhouse cluster.

kurnoolsaketh:
# https://github.com/ClickHouse/clickhouse-kafka-connect/issues/643 PR 2/3
## Summary

Add docker compose clusters setup (3-node ClickHouse with Keeper) and associated code components to enable testing against local clickhouse cluster. Two clusters are defined and in the future, tests will run against both:
- `three_shards_one_replica_each`: to test the connector writing to a multi-shard cluster
- `one_shard_three_replicas`: to test the connector writing to a replicated cluster

the cluster topologies are modeled off the dbt connector: https://github.com/ClickHouse/dbt-clickhouse/blob/main/tests/integration/test_config.xml

### NOTE: this PR is a no-op - it does not rewrite the tests to run against cluster yet. That will be included in the next PR.

closes: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/715


build.gradle.kts
```diff
@@ -334,6 +335,31 @@ task("testJar", type = Jar::class) {
     from(sourceSets.testFixtures.get().allSource)
 }
 
+dockerCompose {
+    useComposeFiles.set(listOf("src/testFixtures/docker/cluster/docker-compose.yml"))
```
chernser: I think we need to explain what cluster and what is actually in this docker compose

src/testFixtures/docker/clickhouse/cluster/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+services:
+  ch0:
+    image: clickhouse/clickhouse-server:${CH_VERSION:-latest}
+    environment:
+      - SERVER_INDEX=1
+      - SHARD_NUM=${SHARD_NUM:-1}
+      - REPLICA_NUM=${REPLICA_NUM:-1}
+      - CLICKHOUSE_SKIP_USER_SETUP=1
+    ports:
+      - "10723:8123"
```
chernser, line 26: I assume this is to expose a single node to connector? 
But in this case we do not test cases when inserts go to different replicas. 

Here how it solved in Metabase tests https://github.com/metabase/metabase/blob/master/modules/drivers/clickhouse/docker-compose.yml

src/testFixtures/docker/clickhouse/cluster/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      - "10743:8443"
+      - "10900:9000"
+    <<: *ch-common
+  ch1:
+    image: clickhouse/clickhouse-server:${CH_VERSION:-latest}
+    environment:
+      - SERVER_INDEX=2
+      - SHARD_NUM=${SHARD_NUM:-2}
```
chernser: Am I correct that if I set `SHARD_NUM=100` before running docker compose it will be places in all shards? In this case may be leave a hardcoded numbers?

src/testFixtures/docker/clickhouse/cluster/config.xml
```diff
@@ -0,0 +1,90 @@
...
+            <shard>
+                <replica>
+                    <host>ch2</host>
+                    <port>9000</port>
+                </replica>
+            </shard>
+        </three_shards_one_replica_each>
+        <one_shard_three_replicas>
+            <shard>
+                <replica>
```
kurnoolsaketh, line 35: set `<internal_replication>true</internal_replication>` here

src/testFixtures/docker/clickhouse/cluster/docker-compose.yml
kurnoolsaketh: good call, thanks @chernser - i've added nginx in front of the cluster

src/testFixtures/docker/clickhouse/cluster/docker-compose.yml
kurnoolsaketh: you are correct, changed these to be hardcoded numbers

chernser (approved):

src/testFixtures/docker/clickhouse/cluster/docker-compose.yml
kurnoolsaketh: @chernser on second thought, this may require some more discussion. there are two ways to write to a sharded cluster (reference: https://clickhouse.com/docs/best-practices/selecting-an-insert-strategy#choose-the-right-insert-target): 

```
- Insert directly into a MergeTree or ReplicatedMergeTree table. This is the most efficient option when the client can perform load balancing across shards. With internal_replication = true, ClickHouse handles replication transparently.
- Insert into a [Distributed table](https://clickhouse.com/docs/engines/table-engines/special/distributed). This allows clients to send data to any node and let ClickHouse forward it to the correct shard. This is simpler but slightly less performant due to the extra forwarding step. internal_replication = true is still recommended.
```

I tried the first approach with nginx, but from what i can tell this introduces a bit more complexity: since we need to wait for writes to be fully replicated before we read, in some cases the proxy marks a node as unresponsive when it's actually just processing a large query backlog and fails future queries on that node. I haven't figured out yet how to tune the timeouts/error tolerance to get this to work for the tests. 

Using the Distributed table might be slightly more straightforward, since there's an option to disable async writes to ensure consistent reads: https://clickhouse.com/docs/operations/settings/settings#distributed_foreground_insert

kurnoolsaketh: @mzitnik what do you think here?

src/testFixtures/docker/clickhouse/cluster/docker-compose.yml
```diff
@@ -0,0 +1,64 @@
+---
+x-ch-common: &ch-common
+#  TODO: add different settings files for different CH versions as necessary (see dbt-clickhouse for reference)
```
mzitnik: maybe a link will be good here?

build.gradle.kts
```diff
@@ -334,6 +335,31 @@ task("testJar", type = Jar::class) {
     from(sourceSets.testFixtures.get().allSource)
 }
 
+dockerCompose {
```
mzitnik: are we starting the cluster from our gradle?

src/testFixtures/docker/clickhouse/cluster/docker-compose.yml
mzitnik: To solve the waiting for replication to be over, we can use [insert_quorum](https://clickhouse.com/docs/operations/settings/settings#insert_quorum)

build.gradle.kts
kurnoolsaketh: discussed offline: I have replaced cluster setup via gradle with java testcontainer. This way, each test file creates its own cluster and we can debug cluster setup in a single JVM instance.

src/testFixtures/docker/clickhouse/cluster/docker-compose.yml
kurnoolsaketh: done

mzitnik (approved):

src/testFixtures/docker/clickhouse/cluster/docker-compose.yml
kurnoolsaketh: decided offline: will proceed with client-side LB (nginx) - it's not as complex as I initially suggested. A follow up PR to this one will will show the usage pattern.

PR #725, opened 2026-04-17
Added cache by Schema

### LINKED ISSUE #718: KeyToValue SMT schema caching bug causes DataException with schema evolution
## Summary

The `KeyToValue` SMT (`com.clickhouse.kafka.connect.transforms.KeyToValue`) caches the schema from the first record it processes and reuses it for all subsequent records. When a topic has multiple Avro schema versions registered in Schema Registry, records written with a different schema version than the cached one fail with `DataException: Struct schemas do not match` during the `Struct.put()` call in `applyWithSchema()`.

## Bug Location

[`KeyToValue.java` — `applyWithSchema()` method](https://github.com/ClickHouse/clickhouse-kafka-connect/blob/main/src/main/java/com/clickhouse/kafka/connect/transforms/KeyToValue.java#L50-L64)

```java
private Schema valueSchema;  // single cached schema

private R applyWithSchema(R record) {
    final Struct oldValue = (Struct) record.value();

    if (valueSchema == null) {  // only built once, never invalidated
        final SchemaBuilder builder = SchemaBuilder.struct();
        // ... builds schema from first record ...
        valueSchema = builder.build();
    }

    Struct newValue = new Struct(valueSchema);
    valueSchema.fields().forEach(f -> {
        if (f.name().equals(keyFieldName)) {
            newValue.put(f, record.key());
        } else {
            newValue.put(f, oldValue.get(f));  // fails if oldValue has a different schema
        }
    });
}
```

The `valueSchema` is set once from the first record and never updated. When a subsequent record arrives with an evolved schema (added/removed fields), `Struct.put()` fails because the cached schema doesn't match the record's actual schema.

## Reproduction

1. Create a topic with 2 Avro schema versions (e.g., v1 with 3 fields, v2 with an additional field in a nested record)
2. Produce messages using both schema versions
3. Create a sink connector using the `KeyToValue` SMT
4. **Result:** Records matching the first-cached schema version succeed; records with the other schema version fail with `DataException` and are routed to DLQ
5. Pause/resume the connector (clears the cache via `close()`) — the behavior flips depending on which schema version is cached first

## Stack Trace

```
org.apache.kafka.connect.errors.DataException: Struct schemas do not match.
    at org.apache.kafka.connect.data.ConnectSchema.validateValue(ConnectSchema.java:253)
    at org.apache.kafka.connect.data.ConnectSchema.validateValue(ConnectSchema.java:217)
    at org.apache.kafka.connect.data.Struct.put(Struct.java:216)
    at com.clickhouse.kafka.connect.transforms.KeyToValue.lambda$applyWithSchema$2(KeyToValue.java:71)
```

chernser:
## Summary
- Implemented schema cache using `Map<Schema, Schema>`. Using schema is key is possible because implementation overrides `hasCode` and `equals` 
- Added a benchmark for testing to verify that old and new implementation are aligned.

Base line:
image

New implementation:
image

Closes: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/718

## Checklist


src/test/java/com/clickhouse/kafka/connect/transforms/KeyToValueTest.java
BentsiLeviav, line 1: I know it has been here for 2 years, we might consider moving the tests to match the source structure `src/test/java/com/clickhouse/kafka/connect/transforms/KeyToValueTest.java`

src/test/java/com/clickhouse/kafka/connect/transforms/KeyToValueTest.java
```diff
@@ -40,6 +40,35 @@ public void applyWithSchemaTest() {
 
 
 
+    @Test
+    public void applyWithDifferentSchemasTest() {
```
BentsiLeviav, line 53: The issue mentioned: `When a topic has multiple Avro schema versions registered in Schema Registry...` and this test uses two completely different schemas on different topics. Can we introduce a test for a schema evolution on the same topic?

benchmark/src/main/java/kafka_connector/KeyToValueBenchmark.java
BentsiLeviav, line 1: All records are built with the same schema. After the first record, any other record just hits the LRU cache. Given that this issue is about schema evolution, do you think it would be worth checking for cache misses or for switching schemas?

src/test/java/com/clickhouse/kafka/connect/transforms/KeyToValueTest.java
chernser: yes, sure.

benchmark/src/main/java/kafka_connector/KeyToValueBenchmark.java
chernser: Do you mean add some debug logging?

chernser:
@claude[agent] review

benchmark/src/main/java/kafka_connector/KeyToValueBenchmark.java
chernser: I've added similar test, too

BentsiLeviav:
@chernser 
The cache key is based only on the value schema, right? Should it also include the key type? We might have issues if, for instance, the key type has changed, and the schema is still the same - with the existing implementation, we will get the same cached schema, because the key is only based on the original schema.

Regardless of whether this is already being taken care of, can we add such a test?

BentsiLeviav (changes requested):

chernser:
@BentsiLeviav 
the schema is key because its equals checks everything: 
```
    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        ConnectSchema schema = (ConnectSchema) o;
        return Objects.equals(optional, schema.optional) &&
                Objects.equals(version, schema.version) &&
                Objects.equals(name, schema.name) &&
                Objects.equals(doc, schema.doc) &&
                Objects.equals(type, schema.type) &&
                Objects.deepEquals(defaultValue, schema.defaultValue) &&
                Objects.equals(fields, schema.fields) &&
                Objects.equals(keySchema, schema.keySchema) &&
                Objects.equals(valueSchema, schema.valueSchema) &&
                Objects.equals(parameters, schema.parameters);
    }
```

it takes into account version and type.

chernser:
@BentsiLeviav ok will do the test.

chernser:
@BentsiLeviav done. I've added test and CacheKey for this case.

src/main/java/com/clickhouse/kafka/connect/transforms/KeyToValue.java
```diff
@@ -23,10 +24,34 @@ public class KeyToValue<R extends ConnectRecord<R>> implements Transformation<R>
                     "Field name on the record value to extract the record key into.");
 
     private String keyFieldName;
-    private Cache<Schema, Schema> schemaUpdateCache;
+    private Cache<CacheKey, Schema> schemaUpdateCache;
 
     protected AtomicInteger cacheMisses = new AtomicInteger(0);
 
+    private static class CacheKey {
```
BentsiLeviav, line 40: What CacheKey does today:
  - Holds two fields (valueSchema, keySchema)
  - Has a constructor that assigns them
  - Has equals that compares both fields
  - Has hashCode based on both fields

Should we transition that to a record?

Something like 
`private record CacheKey(Schema valueSchema, Schema keySchema) {}`

src/main/java/com/clickhouse/kafka/connect/transforms/KeyToValue.java
```diff
@@ -11,19 +14,49 @@
...
+        public int hashCode() {
+            return Objects.hash(valueSchema, keySchema);
+        }
+    }
 
     @Override
     public void configure(Map<String, ?> configs) {
         final SimpleConfig config = new SimpleConfig(CONFIG_DEF, configs);
         keyFieldName = config.getString("field");
+        schemaUpdateCache = new SynchronizedCache<>(new LRUCache<>(16));
```
BentsiLeviav: A small comment - before we had only key as `valueSchema`, so with a cache size of 16, you could remember 16 different value schemas.

Now the cache key is a pair of `(valueSchema, keySchema)`, so now the same value schema appearing with 3 different key schemas takes up 3 slots, not 1.

Ideally, we should make this number configurable, but as a start, do you think it is worth increasing that from 16?

src/main/java/com/clickhouse/kafka/connect/transforms/KeyToValue.java
chernser: Using `record` requires building with java 16+. We have not done the transition yet.

src/main/java/com/clickhouse/kafka/connect/transforms/KeyToValue.java
chernser: In general, changing key structure is not good at all. More like a breaking change. So I would not expect many different schema in the cache. 

I've made size configurable anyway.

BentsiLeviav (approved):

kurnoolsaketh (approved):

PR #736, opened 2026-05-06
[FEAT] support exactly-once with buffer via strict chunking

gaurav7261:
# PR: Exactly-Once Support With Internal Buffer (Strict Chunking)

## Summary

Enables `bufferCount > 0` together with `exactlyOnce = true`. Previously these
were rejected at startup because flush boundaries depended on poll arrival
timing, breaking the determinism that ClickHouse `insert_deduplication_token`
and the KeeperMap-backed state machine require.

This PR adds a **strict-chunking** flush mode that runs only when both flags
are on, leaves the existing at-least-once buffer behavior untouched, and
feeds the same deterministic chunk shape into the existing
`Processing.doLogic` state machine.

## Motivation

Production users with low per-poll throughput (or high throughput restricted
by broker fetch limits) need internal buffering to amortize ClickHouse
insert overhead across multiple `poll()` calls. Until now, choosing
buffering meant choosing at-least-once, even though the rest of the
connector (KeeperMap state, dedup token, OVER_LAPPING split logic) was
perfectly capable of handling exactly-once delivery.

PR #658 shipped buffering for at-least-once and explicitly threw on
`exactlyOnce = true` because the original flush logic violated batch-boundary
determinism in three ways:

1. **Time-based flush** — wall clock varies across retries.
2. **Whole-buffer flush** — overshoot beyond `bufferCount` depends on the
   size of the last poll. Different overshoot on retry → different
   `(minOffset, maxOffset)` → different dedup token.
3. **Cross-partition pooling** — global threshold meant a single partition
   could trigger flushes of records from any other partition with no fixed
   per-partition boundary.

This PR removes all three sources of non-determinism in EO mode.

### Two flush modes

| Mode | Active when | Buffer struct | Threshold | Flush size |
|---|---|---|---|---|
| Relaxed (existing) | `bufferCount > 0` AND `exactlyOnce = false` | flat `List<SinkRecord>` | total record count, OR time | whole buffer |
| Strict chunking (new) | `bufferCount > 0` AND `exactlyOnce = true` | `Map<TopicPartition, List<SinkRecord>>` | per-partition record count | exactly `bufferCount` per chunk |

### Strict-chunking algorithm

```java
// per call to put()
for (SinkRecord r : records) {
    perPartitionBuffer.computeIfAbsent(tp, k -> new ArrayList<>(bufferCount)).add(r);
}
for (List<SinkRecord> bucket : perPartitionBuffer.values()) {
    while (bucket.size() >= bufferCount) {
        List<SinkRecord> head = bucket.subList(0, bufferCount);
        List<SinkRecord> chunk = new ArrayList<>(head);
        head.clear();
        flushChunk(chunk);   // → proxySinkTask.put → Processing.doLogic
    }
}
```

Tail records (`bucket.size() < bufferCount`) remain buffered across `put()`
calls until subsequent calls grow the bucket past `bufferCount`. This is
the determinism guarantee: every flushed chunk has a fixed record count, so
its `(minOffset, maxOffset)` is reproducible on retry given Kafka's
per-offset byte determinism.

### State machine unchanged

The strict-chunking flush feeds chunks into the existing pipeline:

```
flushChunk(chunk)
  → putDirect(chunk)
    → proxySinkTask.put(chunk)
      → Processing.doLogic(records)                ← UNCHANGED
        → stateProvider.getStateRecord(topic, partition)
        → switch(state)         { NONE / BEFORE_PROCESSING / AFTER_PROCESSING }
          switch(overlapState)  { SAME / NEW / OVER_LAPPING / CONTAINS /
                                  ZERO / PREVIOUS / ERROR }
        → dbWriter.doInsert(records, queryId)
              ↑ QueryIdentifier.getDeduplicationToken()
                = "topic-partition-minOffset-maxOffset"
```

All EO guarantees flow from this layer:

- **Dedup token** carries `(topic, partition, min, max)` — reproducible
  because chunk boundaries are now reproducible.
- **KeeperMap state** records the last `(min, max)` per partition with
  `BEFORE_PROCESSING`/`AFTER_PROCESSING` markers, fencing both partial-write
  failures and replays beyond the dedup window.
- **`OVER_LAPPING` branch** splits incoming batches against the stored
  range, dropping the already-processed prefix and inserting only the
  truly-new suffix — this absorbs minor boundary divergence as well.

### Validation

`ClickHouseSinkTask.start()` rejects unsafe combinations via
`validateBufferConfig()`:

| `bufferCount` | `exactlyOnce` | `bufferFlushTime` | `ignorePartitionsWhenBatching` | Result |
|---|---|---|---|---|
| 0 | any | any | any | no buffering, original path |
| > 0 | false | any | any | relaxed mode (existing) |
| > 0 | true | 0 | false | **strict-chunking (new)** |
| > 0 | true | > 0 | any | `ConnectException` |
| > 0 | true | 0 | true | `ConnectException` |

Reasons for the throws:

- `bufferFlushTime > 0` introduces wall-clock-driven flushes that cannot
  be reproduced on retry.
- `ignorePartitionsWhenBatching = true` makes
  `QueryIdentifier.getDeduplicationToken()` return `null` (partition is set
  to `-1`), disabling token-based dedup entirely.

## Configuration Recipe

For users wanting EO + buffer:

```jsonc
{
  "exactlyOnce": "true",
  "ignorePartitionsWhenBatching": "false",
  "bufferCount": "60000",
  "bufferFlushTime": "0",

  "tasks.max": "<= partition count>",
  "consumer.override.max.poll.records": "<less than bufferCount>",

  "clickhouseSettings": "insert_deduplicate=1,replicated_deduplication_window=1000"
}
```

Notes:

- `tasks.max ≤ partition count` keeps assignor stable and prevents
  unnecessary cross-task fanout per partition.
- `consumer.override.max.poll.records < bufferCount` ensures buffering
  actually amortizes — otherwise each poll fills a chunk by itself.
- `replicated_deduplication_window` should exceed the worst-case replay
  window in chunks. Default `100` is small for high-throughput sinks.
- The existing pipeline assumes deterministic record content. Drop SMTs
  that introduce non-determinism per replay (e.g. custom transforms
  injecting `currentTimeMillis()` or `UUID.randomUUID()`). Stock
  `InsertField$Value` with `timestamp.field`, `topic.field`, `offset.field`,
  `partition.field`, and `TimestampConverter$Value` are deterministic
  because they read fields already baked into the Kafka record.

### Validator tests

- `exactlyOnceWithBufferRejectsTimeFlush` — `bufferFlushTime > 0` throws.
- `exactlyOnceWithBufferRejectsIgnorePartitions` — `ignorePartitionsWhenBatching = true` throws.

### Strict-chunking behavior

- `strictChunkingStartsWithValidConfig` — happy-path startup.
- `strictChunkingTailRecordsRemainBuffered` — tail < threshold stays
  buffered; verified with `assertCountStaysAt()` polling for 2s to guard
  against delayed-write false negatives.
- `strictChunkingFlushesNRecordChunks` (parametrized, 6 cases) — covers
  below-threshold (499), exact threshold (500), single-chunk overshoot
  (700), exact two chunks (1000), two chunks + tail (1100), exact three
  chunks (1500). Verifies both the strict-N flush size and the tail-stays-buffered
  invariant.
- `strictChunkingPerPartitionIndependent` — P0 hits threshold, P1 does
  not, only P0 flushes.
- `strictChunkingTailNotFlushedOnStop` — `stop()` discards tail
  (Kafka redelivers on restart).
- `strictChunkingPreCommitOnlyForFlushedChunks` — `preCommit()` returns
  the flushed-chunk offset, not the tail offset.
- `strictChunkingRebalanceDropsPartitionBucket` — `close([P1])` removes
  P1's bucket and `flushedOffsets` entry.

### Existing tests

All relaxed-mode tests preserved unchanged. They continue to exercise
the flat-buffer code path for `exactlyOnce = false` users.

### Test infrastructure

Standalone CH testcontainer mounts `src/test/resources/clickhouse_keeper_map.xml`
into `/etc/clickhouse-server/config.d/`. Enables embedded `clickhouse-keeper`
on port 9181 and sets `keeper_map_path_prefix`, so the KeeperMap-backed
`connect_state` table required by EO tests can be created in standalone
runs. Result: full EO test coverage on every CI matrix entry, not just
the cloud workflow.

Local run on standalone CH 25.8 + podman: 22 single tests + 6 parametrized
cases = 28 PASSED, build clean in 25s.

## Backward Compatibility

- Relaxed-mode users (`exactlyOnce = false`): zero behavior change.
- EO users without buffering (`bufferCount = 0`): zero behavior change.
- EO users who previously hit the `ConnectException` thrown by PR #658
  can now opt in to strict chunking by setting `bufferFlushTime = 0` and
  `ignorePartitionsWhenBatching = false`. No silent semantic change —
  any unsafe combination still throws at startup with a specific
  diagnostic.

## Code review changes (post-review)

Address @chernser's review comments:

- Extracted `validateBufferConfig()` static method, removed redundant
  local variables, inlined `isExactlyOnce()` at single use site.
- Extracted `head` variable in `putStrict` so `subList(0, bufferCount)`
  evaluates once.
- Added Javadoc on `flushChunk()` documenting failure semantics for
  no-tolerance throw, DLQ-tolerated path, and JVM-crash between
  `head.clear()` and `putDirect()`.
- Expanded `exactlyOnceWithBufferRejectsIgnorePartitions` comment with
  exact line references to `QueryIdentifier.java:14-22, 56-61`.
- Added `assertCountStaysAt()` helper polling every 250ms for 2s to guard
  zero-count assertions against delayed-write false negatives.
- Merged two single-input strict-chunking tests into a parametrized
  `@CsvSource` test with 6 boundary cases.

## Out of Scope / Known Limitations

- **Non-deterministic SMTs** still defeat content-level dedup (e.g.
  custom transforms calling `System.currentTimeMillis()` or
  `UUID.randomUUID()` per record). Token-level dedup, the primary
  mechanism here, is unaffected because the token is computed from
  Kafka offsets, not record content. Stock `InsertField$Value` with
  `timestamp.field` reads `record.timestamp()` — already part of the
  immutable Kafka record — so it is deterministic on replay.
- **Tail latency**: in strict-chunking mode, slow partitions can hold
  records below threshold for an unbounded time. Users needing bounded
  latency should size `bufferCount` against expected steady-state
  throughput, or disable buffering entirely (each `put()` then flushes
  its natural batch via the no-buffer path).
- **CH dedup window sizing**: long replay windows may exhaust the default
  100-block dedup memory. Bump `replicated_deduplication_window` and
  `replicated_deduplication_window_seconds` accordingly.
- **No 2PC**: ClickHouse cannot participate in true two-phase commit
  (no durable PREPARED state across sessions). EO continues to be
  achieved via idempotent retry — token reuse plus state-machine
  reconciliation. This PR does not change that fundamental.

## Follow-up items (not in this PR)

- Single `deliveryMode` enum (`AT_LEAST_ONCE` / `AT_LEAST_ONCE_BUFFERED`
  / `EXACTLY_ONCE` / `EXACTLY_ONCE_BUFFERED`) replacing the current
  flag matrix. Suggested by @chernser in review; deferred because it is
  a public-API change requiring deprecation handling for the existing
  flags. Tracked separately.
- Optional offset-aligned tail-flush escape hatch (e.g.
  `bufferStrictTailMaxAge`) for very-low-throughput topics that need
  bounded latency in EO mode. State machine's `OVER_LAPPING` branch
  should absorb the resulting boundary variance.


gaurav7261:
@mzitnik  checked logs, do cloud does not have KeeperMap table or should i exclude cloud runs on tests 🤔

chernser:
@gaurav7261 

There are multiple tests fail with: 
```
ClickHouseSinkTaskBufferTest > strictChunkingDrainsMultipleChunksInOnePut() FAILED
    java.lang.RuntimeException: com.clickhouse.client.ClickHouseException: Code: 36. DB::Exception: KeeperMap is disabled because 'keeper_map_path_prefix' config is not defined. (BAD_ARGUMENTS) (version 25.8.22.28 (official build))
        at com.clickhouse.kafka.connect.sink.db.helper.ClickHouseHelperClient.queryV1(ClickHouseHelperClient.java:284)
        at com.clickhouse.kafka.connect.sink.db.helper.ClickHouseHelperClient.queryV1(ClickHouseHelperClient.java:262)
        at com.clickhouse.kafka.connect.sink.state.provider.KeeperStateProvider.init(KeeperStateProvider.java:89)
        at com.clickhouse.kafka.connect.sink.state.provider.KeeperStateProvider.<init>(KeeperStateProvider.java:64)
        at com.clickhouse.kafka.connect.sink.ProxySinkTask.<init>(ProxySinkTask.java:47)
        at com.clickhouse.kafka.connect.sink.ClickHouseSinkTask.start(ClickHouseSinkTask.java:84)
        at com.clickhouse.kafka.connect.sink.ClickHouseSinkTaskBufferTest.strictChunkingDrainsMultipleChunksInOnePut(ClickHouseSinkTaskBufferTest.java:620)

        Caused by:
        com.clickhouse.client.ClickHouseException: Code: 36. DB::Exception: KeeperMap is disabled because 'keeper_map_path_prefix' config is not defined. (BAD_ARGUMENTS) (version 25.8.22.28 (official build))
            at app//com.clickhouse.client.ClickHouseException.of(ClickHouseException.java:152)
            at app//com.clickhouse.client.AbstractClient.lambda$execute$0(AbstractClient.java:282)
            at java.base@17.0.18/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
            at java.base@17.0.18/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
            at java.base@17.0.18/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
            at java.base@17.0.18/java.lang.Thread.run(Thread.java:840)

            Caused by:
            java.io.IOException: Code: 36. DB::Exception: KeeperMap is disabled because 'keeper_map_path_prefix' config is not defined. (BAD_ARGUMENTS) (version 25.8.22.28 (official build))
                at com.clickhouse.client.http.ApacheHttpConnectionImpl.checkResponse(ApacheHttpConnectionImpl.java:262)
                at com.clickhouse.client.http.ApacheHttpConnectionImpl.post(ApacheHttpConnectionImpl.java:325)
                at com.clickhouse.client.http.ClickHouseHttpClient.send(ClickHouseHttpClient.java:196)
                at com.clickhouse.client.AbstractClient.sendAsync(AbstractClient.java:165)
                at com.clickhouse.client.AbstractClient.lambda$execute$0(AbstractClient.java:280)
                ... 4 more
```

Please make sure to run tests locally by `./gradlew test` . Thanks!

gaurav7261:
Hi @chernser ,seems like all were of cloud only,let me check again

chernser:
@gaurav7261 
only cloud tests passing 
Here is what I see - standalone tests failing. Do you run locally?  
image

gaurav7261:
got the issue @chernser , updated, tested locally, checking ci now

chernser:
Thank you, @gaurav7261 ! 
I will review it today.

gaurav7261:
@chernser can you please review it

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
```diff
@@ -221,14 +221,32 @@ public void bufferAccumulatesMultipleBatches() {
...
         ClickHouseSinkTask task = new ClickHouseSinkTask();
         assertThrows(ConnectException.class, () -> task.start(props),
-                "Buffering should not be allowed with exactly-once mode");
+                "EO + buffer must throw when bufferFlushTime > 0");
+    }
+
+    @Test
+    public void exactlyOnceWithBufferRejectsIgnorePartitions() {
+        // EO + buffer requires per-partition batching. ignorePartitionsWhenBatching=true
+        // makes the dedup token's partition component null → CH dedup can't fire.
```
chernser: would you please explain this statement? 
is it another requirement?

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
```diff
@@ -221,14 +221,32 @@ public void bufferAccumulatesMultipleBatches() {
     }
 
     @Test
-    public void bufferIncompatibleWithExactlyOnce() {
+    public void exactlyOnceWithBufferRejectsTimeFlush() {
+        // EO + buffer requires bufferFlushTime=0. Time-based flush is non-deterministic
+        // across retries (wall clock), which breaks dedup-token reuse.
```
chernser, line 229: It feels like we need a delivery mode configuration parameter
- EO
- NOT EO 
- BUFF
- BUFF + EO

chernser: assuming each has own configuration.

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
```diff
@@ -515,4 +533,207 @@ public void bufferMultiplePartitions() {
...
+        new CreateTableStatement(PRIMITIVE_TYPES_TABLE).tableName(topic).execute(chc);
+
+        ClickHouseSinkTask task = new ClickHouseSinkTask();
+        task.start(props);
+
+        // 300 records below 500 threshold → all stay buffered.
+        List<SinkRecord> batch = SchemalessTestData.createPrimitiveTypes(topic, 1, 300);
+        task.put(batch);
+
+        assertEquals(0, ClickHouseTestHelpers.countRows(chc, topic),
```
chernser: there should be a time delay or loop checking for period of time to check that record really missing (they  can be delayed on write)

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
```diff
@@ -515,4 +533,207 @@ public void bufferMultiplePartitions() {
...
+                "Tail records below bufferCount must remain buffered");
+        task.stop();
+    }
+
+    @Test
+    public void strictChunkingFlushesExactlyBufferCountRecords() {
+        // 700 records with bufferCount=500 → flush exactly 500, tail 200 stays buffered.
```
chernser: I think, this is not a right strategy - why do we need leave 200 records hanging in the memory when we can just insert them?

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
```diff
@@ -515,4 +533,207 @@ public void bufferMultiplePartitions() {
...
+        List<SinkRecord> batch = SchemalessTestData.createPrimitiveTypes(topic, 1, 700);
+        task.put(batch);
+
+        assertEquals(500, ClickHouseTestHelpers.countRows(chc, topic),
+                "Strict chunking must flush exactly bufferCount=500, not the overshoot 700");
+        task.stop();
+    }
+
+    @Test
+    public void strictChunkingDrainsMultipleChunksInOnePut() {
```
chernser: please join with previous test - make it parametrized. 
the problem with multiple similar tests - we have to review them and update over time.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -50,31 +57,52 @@ public void start(Map<String, String> props) {
...
         int configBufferCount = clickHouseSinkConfig.getBufferCount();
-        if (configBufferCount > 0 && clickHouseSinkConfig.isExactlyOnce()) {
-            throw new ConnectException(
-                    "Internal buffering (bufferCount > 0) is not compatible with exactly-once mode. " +
-                    "Buffering changes batch boundaries, which breaks ClickHouse block deduplication and the offset state machine. " +
-                    "To resolve this, either disable exactly-once mode by setting 'exactlyOnce=false' in your connector config, " +
-                    "or disable buffering by setting 'bufferCount=0'."
-            );
+        long configBufferFlushTime = clickHouseSinkConfig.getBufferFlushTime();
+        boolean eo = clickHouseSinkConfig.isExactlyOnce();
```
chernser: please use `clickhouseSinkConfig.isExactlyOnce()` inline because it is more clear and needed in one place only

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -50,31 +57,52 @@ public void start(Map<String, String> props) {
             throw new ConnectException("Failed to start new task" , e);
         }
 
-        // Validate config before heavy initialization
+        // Validate buffer + exactly-once compatibility. EO with buffering is supported
+        // only under strict-chunking constraints that preserve batch-boundary determinism.
         int configBufferCount = clickHouseSinkConfig.getBufferCount();
```
chernser: just a minor: variable names can be shorter - no need to mention config. 
for example 
``` 
int buffers = ...
long flushTime = ..
```

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -50,31 +57,52 @@ public void start(Map<String, String> props) {
             throw new ConnectException("Failed to start new task" , e);
         }
 
-        // Validate config before heavy initialization
+        // Validate buffer + exactly-once compatibility. EO with buffering is supported
```
chernser: please move validation of buffer config to a dedicated method.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -86,42 +114,93 @@ public void put(Collection<SinkRecord> records) {
...
+            }
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Buffered {} records across {} partitions (strict)",
+                        records.size(), perPartitionBuffer.size());
+            }
+        }
+
+        for (List<SinkRecord> bucket : perPartitionBuffer.values()) {
+            while (bucket.size() >= bufferCount) {
+                List<SinkRecord> chunk = new ArrayList<>(bucket.subList(0, bufferCount));
```
chernser: bucket.subList called twice. use a variable?

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -86,42 +114,93 @@ public void put(Collection<SinkRecord> records) {
...
+                bucket.subList(0, bufferCount).clear();
```
chernser: Would you please document what happens if flush fails?

chernser (changes requested):

chernser:
@gaurav7261
I've done review - please see comments

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
gaurav7261: @chernser i am more inclined to do same in another PR, what's your thoughts on this?

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
gaurav7261: fixed @chernser

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
gaurav7261: @chernser Tail isn't lost — it's the determinism guarantee. Flushing 700 makes token t-0-0-699; on retry with 500+200 poll split, token becomes t-0-0-499 → mismatch → dup. Strict N keeps token reproducible across poll groupings. Want a follow-up bufferStrictTailMaxAge escape hatch for low-throughput  topics, or ship strict semantics as v1?

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
gaurav7261: done

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
chernser: I think, it is related to this PR. But lets ask what others think 

@kurnoolsaketh @mzitnik

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -50,31 +57,55 @@ public void start(Map<String, String> props) {
...
         this.proxySinkTask = new ProxySinkTask(clickHouseSinkConfig, errorReporter);
         this.proxySinkTask.start();
 
-        // Initialize buffering
-        this.bufferCount = configBufferCount;
+        this.bufferCount = clickHouseSinkConfig.getBufferCount();
         this.bufferFlushTime = clickHouseSinkConfig.getBufferFlushTime();
         this.bufferingEnabled = this.bufferCount > 0;
-        this.buffer = this.bufferingEnabled ? new ArrayList<>(this.bufferCount) : new ArrayList<>();
+        this.strictChunking = this.bufferingEnabled && clickHouseSinkConfig.isExactlyOnce();
```
chernser: just to point a problem with tangled configuration 
`strictChunking` depends on two other flags. But this knowledge is simple lost once we move away our attention from this method. Later we need to lookup what is strictChunking is and if we need to check if buffering is enabled. 
It is in context of introducing insert modes. `strictChunking` is sub configuration of  buffered insert.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -216,10 +331,21 @@ public void stop() {
         // Note: close() is called before stop() by the framework, which already
         // removes buffered records. Unflushed records' offsets were never committed
         // via preCommit(), so Kafka will redeliver them on restart (at-least-once).
-        if (bufferingEnabled && buffer != null && !buffer.isEmpty()) {
-            LOGGER.warn("Stop called with {} buffered records still in buffer — " +
-                    "these will be redelivered on restart since offsets were not committed", buffer.size());
-            buffer.clear();
+        if (bufferingEnabled) {
+            int remaining = 0;
+            if (strictChunking && perPartitionBuffer != null) {
```
chernser: should `perPartitionBuffer` be checked on like 300 ?

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
chernser: ok, it may be will complicate the algorithm. 
would it be the same case when user reconfigures buffer size? should they be prepared?

chernser (changes requested):

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: done

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: done

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: done

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
gaurav7261: cool waiting for @kurnoolsaketh @mzitnik  review, if needed i can make the necessary changes in this PR only

chernser:
@gaurav7261 
In general looks good. 
But I have not found the test where deduplication is verified.
- we got 200 records then same 200 + 300 - should get only 500 records 
- we got 200 records then another 200 then last 200 + 100 records - should get only 500 records. 


And I've just realized that there can be a case when for long period it could be no records and data will stuck for too long. Need to another timeout for such case (both for strict and relaxed) 

CC: @mzitnik @kurnoolsaketh

gaurav7261:
@chernser added test  Replayed Flush, ReplayWithBufferedTail, AcrossMultipleChunkReplays, can you please review

chernser:
@gaurav7261 would you please also resolve merge conflict so CI can run tests? 
Thanks!

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
```diff
@@ -515,4 +540,337 @@ public void bufferMultiplePartitions() {
...
+    public void strictChunkingStartsWithValidConfig() {
+        // Valid EO + buffer combo must start without throwing.
+        Map<String, String> props = strictChunkingProps(500);
+        ClickHouseHelperClient chc = ClickHouseTestHelpers.createClient(props);
+        String topic = createTopicName("strict_start_test");
+        ClickHouseTestHelpers.dropTable(chc, topic);
+        new CreateTableStatement(PRIMITIVE_TYPES_TABLE).tableName(topic).execute(chc);
+
+        ClickHouseSinkTask task = new ClickHouseSinkTask();
+        task.start(props);
```
chernser: this test verifies that valid configuration is not rejected but test doesn't verify if ingestion really work. 
it should be part of a positive test or unit test. 
Here we should test different valid configuration for expected result.

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
```diff
@@ -515,4 +540,337 @@ public void bufferMultiplePartitions() {
...
+        runOne.stop();
+
+        // Run 2: Kafka redelivers offsets [0, 499]. Buffer hits threshold, attempts flush.
+        // Same offsets → same token → ClickHouse insert_deduplication_token catches it,
+        // and the state machine's AFTER_PROCESSING + SAME branch suppresses the insert
+        // entirely.
+        ClickHouseSinkTask runTwo = new ClickHouseSinkTask();
+        runTwo.start(props);
+        runTwo.put(SchemalessTestData.createPrimitiveTypes(topic, 1, 500));
+        assertEquals(500, ClickHouseTestHelpers.countRows(chc, topic),
```
chernser: may be use `assertCountStaysAt` ?

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
```diff
@@ -515,4 +540,337 @@ public void bufferMultiplePartitions() {
...
+
+        // Run 1: flush 500 records → token topic-1-0-499, state stored AFTER_PROCESSING.
+        ClickHouseSinkTask runOne = new ClickHouseSinkTask();
+        runOne.start(props);
+        runOne.put(SchemalessTestData.createPrimitiveTypes(topic, 1, 500));
+        assertEquals(500, ClickHouseTestHelpers.countRows(chc, topic),
+                "Run 1 should flush exactly 500 records");
+        // Simulate crash: offset commit lost (do not call preCommit). Buffer was already
+        // empty after the strict flush, so stop() drops nothing.
+        runOne.stop();
```
chernser: there is not always stop - we can get duplicates while running task. So you can parametrize test with and without shutdown.

chernser (changes requested):

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
gaurav7261: done

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
gaurav7261: done

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -25,12 +26,39 @@ public class ClickHouseSinkTask extends SinkTask {
...
     private ErrorReporter errorReporter;
 
-    // Internal buffering
+    // Internal buffering.
+    // At-least-once mode: single flat buffer, threshold = total record count, whole-buffer flush.
+    // Strict-chunking mode (exactly-once): per-partition buckets, threshold = per-partition count,
+    //   flushes in fixed bufferCount-sized chunks. Keeps batch boundaries deterministic across
+    //   retries so ClickHouse insert_deduplication_token reuses correctly.
     private List<SinkRecord> buffer;
+    private Map<TopicPartition, List<SinkRecord>> perPartitionBuffer;
```
kurnoolsaketh: I think the current implementation is slightly difficult to read. Why have a different buffer for "strict" versus "relaxed" mode? Why not just use `Map<TopicPartition, List<SinkRecord>> perPartitionBuffer` for both modes? You can probably eliminate the putStrict/putRelaxed forking logic below and just have one codepath for both since the chunking mechanism is the same either way.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -25,12 +26,39 @@ public class ClickHouseSinkTask extends SinkTask {
...
+     * com.clickhouse.kafka.connect.sink.processing.Processing} state-machine reconciliation.
+     *
+     * <p>When false, buffering (if enabled) uses the relaxed flat-buffer path from PR #658.
+     *
+     * <p>Note: this flag is bookkeeping for the boolean combination of
+     * {@code bufferCount > 0} + {@code exactlyOnce = true}. If the public configuration
+     * surface is later replaced by a single {@code deliveryMode} enum, this field collapses
+     * into the enum's {@code EXACTLY_ONCE_BUFFERED} variant.
+     */
+    private boolean strictChunking;
```
kurnoolsaketh: strictChunking is a new terminology you are introducing in this PR - imo, it takes an extra mental step to understand that this is referring to exactly-once + buffering enabled. I would prefer if you renamed this flag to exactly what it represents, ie something like `exactlyOnceAndBufferingEnabled`.

this way, you can also remove the verbose comment here - imo, this is the wrong place for these sorts of design descriptions.

kurnoolsaketh: please add design/architecture descriptions to DESIGN.md

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -86,42 +161,117 @@ public void put(Collection<SinkRecord> records) {
...
+        } else {
+            putRelaxed(records);
+        }
+    }
+
+    /**
+     * At-least-once buffering. Single flat buffer, threshold on total size,
+     * whole-buffer flush. Preserves the semantics shipped in PR #658.
+     */
+    private void putRelaxed(Collection<SinkRecord> records) {
```
kurnoolsaketh: please rename this to `putBufferAtLeastOnce`

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -86,42 +161,117 @@ public void put(Collection<SinkRecord> records) {
...
 
+    /**
+     * Exactly-once buffering. Buckets per partition, drains in fixed
+     * bufferCount-sized chunks. Tail records below threshold remain buffered
+     * across put() calls until a subsequent call pushes the bucket over
+     * bufferCount. Each flush has a fixed record count, so (minOffset, maxOffset)
+     * is reproducible across retries — required for ClickHouse
+     * insert_deduplication_token reuse and StateProvider range comparison.
+     */
+    private void putStrict(Collection<SinkRecord> records) {
```
kurnoolsaketh: please rename this to putBufferExactlyOnce

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -86,42 +161,117 @@ public void put(Collection<SinkRecord> records) {
...
     private void flushBuffer() {
         if (buffer.isEmpty()) {
             return;
         }
         List<SinkRecord> toFlush = new ArrayList<>(buffer);
-        putDirect(toFlush);
-        // Track the max flushed offset per topic/partition so preCommit() only
-        // allows the framework to commit offsets for data written to ClickHouse
-        for (SinkRecord record : toFlush) {
+        buffer.clear();
```
kurnoolsaketh: why are we copying the buffer here instead of just clearing it after flushChunk returns? If i'm not mistaken, flushChunk does not mutate this list, so defensive copying is not necessary here.

kurnoolsaketh: actually, if you remove the forking logic, this function is not necessary anymore.

kurnoolsaketh (commented):
thanks @gaurav7261 for the contribution. looks mostly sensible, but naming needs work and i think the putStrict/putRelaxed forking logic is not necessary. I will take another pass once you address this round of reviews.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: done

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: done

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: done

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: @kurnoolsaketh  One thing I want to confirm here: collapsing to a single per-partition codepath changes the at-least-once flush semantic that PR #658 shipped:

 - PR #658 today: global threshold across all partitions (buffer.size() >= bufferCount), single whole-buffer flush.
 - Unified per-partition: each partition has its own bufferCount threshold; flushes happen partition-by-partition.

 For users with traffic spread across many partitions, the unified version would produce more, smaller CH inserts at the same bufferCount setting (a partition rarely hits the global threshold alone), so  they'd need to tune up bufferCount to get the same coalescing. That's a behavioral break for relaxed-mode users.

 I can think of two ways:
 1. Soft unification — share the Map<TopicPartition, List<SinkRecord>> data structure but keep two flush strategies (whole-buffer for at-least-once, strict-N for exactly-once). It will eliminates the duplicate  field but keeps both semantics. 
 2. Hard unification — drop the global-threshold semantic entirely but this will break my PR #658.

kurnoolsaketh: ok thanks, it's fine as is to me. The putBufferExactlyOnce/putBufferAtLeastOnce naming change makes it clear enough 👍 . we can close this thread

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
kurnoolsaketh, line 1: @gaurav7261 in addition to this, please also add integration tests for the buffering + EO config to `src/integrationTest/java/com/clickhouse/kafka/connect/sink/ExactlyOnceTest.java`. You should:
- create a new `.json` sink config with the buffering + EO config (you can simply copy src/integrationTest/resources/clickhouse_sink_no_proxy.json and src/integrationTest/resources/clickhouse_sink_no_proxy_schemaless.json). So, something like `clickhouse_sink_no_proxy_buffer.json` and `clickhouse_sink_no_proxy_schemaless_buffer.json`
- make checkTotalsEqual, checkTotalsEqualMulti, checkSpottyNetwork, and checkSpottyNetworkMulti parameterized on the config filename. Ie, change `@Test` to @ParameterizedTest("getConfigs") and getConfigs() returns either the 2 schemaless configs or the 2 with-schema configs.

kurnoolsaketh (changes requested):

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
gaurav7261: great thanks @kurnoolsaketh

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskBufferTest.java
gaurav7261: @kurnoolsaketh done, can you please review

gaurav7261:
@kurnoolsaketh @chernser can we merge this now ?

src/integrationTest/java/com/clickhouse/kafka/connect/sink/ExactlyOnceTest.java
kurnoolsaketh: hi @gaurav7261, please create `clickhouse_sink_no_proxy_buffer.json` and add `clickhouse_sink_no_proxy.json` and` clickhouse_sink_no_proxy_buffer.json` to `getConfigs` as well

kurnoolsaketh (changes requested):

gaurav7261:
@kurnoolsaketh please review now cc @mzitnik

gaurav7261:
@kurnoolsaketh all test passed, can you please review cc @mzitnik

chernser (approved):

mzitnik:
@gaurav7261 looking into it

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -86,42 +136,115 @@ public void put(Collection<SinkRecord> records) {
...
+     * across put() calls until a subsequent call pushes the bucket over
+     * bufferCount. Each flush has a fixed record count, so (minOffset, maxOffset)
+     * is reproducible across retries — required for ClickHouse
+     * insert_deduplication_token reuse and StateProvider range comparison.
+     */
+    private void putBufferExactlyOnce(Collection<SinkRecord> records) {
+        if (records != null && !records.isEmpty()) {
+            for (SinkRecord r : records) {
+                TopicPartition tp = new TopicPartition(r.topic(), r.kafkaPartition());
+                perPartitionBuffer.computeIfAbsent(tp, k -> new ArrayList<>(bufferCount)).add(r);
```
mzitnik: `bufferCount` acts here as a global count, so we overallocate space by doing `new ArrayList<>(bufferCount))`

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTask.java
```diff
@@ -86,42 +136,115 @@ public void put(Collection<SinkRecord> records) {
...
+            }
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Buffered {} records across {} partitions (strict)",
+                        records.size(), perPartitionBuffer.size());
+            }
+        }
+
+        for (List<SinkRecord> bucket : perPartitionBuffer.values()) {
+            while (bucket.size() >= bufferCount) {
+                List<SinkRecord> head = bucket.subList(0, bufferCount);
```
mzitnik: Something is not clear about what happens 
- if bucket.size() < bufferCount
- bucket.subList(0, bufferCount); we have the tail

mzitnik (commented):
Started to review some things that are not clear. Pinging in our community Slack

gaurav7261:
@mzitnik can you please review it

chernser (approved):

PR #756, opened 2026-05-26
[FIX] bound writer V2 client insert so hung ClickHouse cannot stall SinkTask

gaurav7261:
# Add bounded timeout to V2 client insert calls

## Summary

`ClickHouseWriter` invokes `client.insert(...).get()` (Java client V2) with no timeout at three call sites. If ClickHouse stalls (network partition, server hang, GC pause, slow merges) the SinkTask thread blocks indefinitely. Once the block exceeds `consumer.override.max.poll.interval.ms`, Kafka treats the worker as dead and triggers a consumer-group rebalance — partitions get reassigned mid-insert, observability gets noisy, and recovery is slow.

This PR adds a bounded `Future.get(timeout, TimeUnit)` to all three V2 insert paths and surfaces the timeout as a `RuntimeException`, so existing `errors.tolerance` / DLQ / state-machine retry logic handles it the same way it handles any other transient insert failure.

### New configuration

| Key | Type | Default | Notes |
|-----|------|---------|-------|
| `clickhouseClientInsertTimeoutMs` | long (>= 1000) | `240000` (4 min) | Maximum wait per V2 insert response. Should be set below `consumer.override.max.poll.interval.ms`. |

A startup validator emits a `WARN` if `clickhouseClientInsertTimeoutMs >= consumer.override.max.poll.interval.ms`.

## Failure semantics

| Mode | Timeout behaviour |
|------|------------------|
| At-least-once (`exactlyOnce=false`) | `RuntimeException` propagates → `Utils.handleException` → DLQ if `errors.tolerance=all`, else task fails → Kafka redelivers the offset range on next poll |
| Exactly-once (`exactlyOnce=true`) | `RuntimeException` propagates → state stays in `BEFORE_PROCESSING` → on retry `Processing.doLogic` reconstructs the **byte-for-byte identical** block (same offsets, deterministic ordering via strict chunking, same `insert_deduplication_token`) → ClickHouse drops the duplicate via `replicated_deduplication_window` if the first attempt actually landed |

The cancel call (`future.cancel(true)`) frees the underlying HTTP connection back to the V2 client pool so subsequent batches do not stack against the hung connection.

## Why a dedicated config (not reusing `timeoutSeconds`)

`timeoutSeconds` already governs short ClickHouse driver operations (DDL, ping, version checks); reusing it for inserts would force everyone to either (a) raise short-op timeouts to 4 min — increasing tail latency on health checks — or (b) lower insert timeouts to 30 s — too aggressive for large EO batches. The two operations have different sensitivity profiles, so they get separate knobs.


gaurav7261:
@chernser @mzitnik @kurnoolsaketh please review

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -1425,6 +1427,37 @@ private ClickHouseClient getClient() {
...
+     *
+     * <p>Failure semantics:
+     * <ul>
+     *   <li>at-least-once: timeout -> exception -> records redelivered by Kafka.</li>
+     *   <li>exactly-once: timeout -> exception -> on retry the same block is reconstructed
+     *       (deterministic offsets + insert_deduplication_token) and ClickHouse drops duplicates.</li>
+     * </ul>
+     */
+    private InsertResponse awaitInsertResponse(Future<InsertResponse> future, QueryIdentifier queryId)
+            throws ExecutionException, InterruptedException {
```
chernser, line 1442: Exceptions  should be handled in this method. Anyway it handles `TimeoutException`.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -1425,6 +1427,37 @@ private ClickHouseClient getClient() {
...
+        long timeoutMs = csc.getClickhouseClientInsertTimeoutMs();
+        try {
+            return future.get(timeoutMs, TimeUnit.MILLISECONDS);
+        } catch (TimeoutException e) {
+            future.cancel(true);
+            String msg = String.format(
```
chernser: First of all, avoid using `String.format` - it is heavy operation and hardly readable construction. There is a good chance to get error if operand is missing. `+` work much better and more readable

And we are using slf4j that understand `{}` (but not int `.error()`). 

In this case log would be a `warn` not error. Because error will be reported later when thrown Exception reaches catch block. 
In general exceptions should be logged on thrown. Both might be too many. In this case good exception message is enough.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
```diff
@@ -680,6 +700,21 @@ private static ConfigDef createConfigDef() {
                 ConfigDef.Width.SHORT,
                 "Buffer flush time in ms."
         );
+        configDef.define(CLICKHOUSE_CLIENT_INSERT_TIMEOUT_MS,
+                ConfigDef.Type.LONG,
+                clickhouseClientInsertTimeoutMsDefault,
+                ConfigDef.Range.atLeast(1000),
```
chernser: it should be around `100 ms`. Writes are fast and we should not limit in this way.

src/main/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkConfig.java
```diff
@@ -680,6 +700,21 @@ private static ConfigDef createConfigDef() {
...
+                ConfigDef.Importance.LOW,
+                "Maximum time in milliseconds to wait for a single ClickHouse insert response before failing the batch. " +
```
chernser: - dependency on `poll.interval.ms` should be first thing to tell
- what is good margin between values?
- we should just explain if retriable or not retriable exception will be thrown.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
chernser: seems I'm wrong about this exceptions - they are caught later.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -1425,6 +1427,37 @@ private ClickHouseClient getClient() {
...
+                    "ClickHouse insert timed out after %d ms (QueryId: [%s])",
+                    timeoutMs, queryId.getQueryId());
+            LOGGER.error(msg);
+            throw new RuntimeException(msg, e);
```
chernser: This will result not retriable exception. However socket timeout, for example, is retriable. Please look into `com.clickhouse.kafka.connect.util.Utils#handleException` and make  this case also retriable.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 1.3.10, 2026-05-26
+
+## Bug Fixes
+* Added bounded timeout for V2 client insert calls. `Future.get()` was unbounded; a hung ClickHouse could stall the SinkTask thread past `consumer.override.max.poll.interval.ms` and trigger Kafka consumer rebalance. New `clickhouseClientInsertTimeoutMs` config (default 240000 ms) caps the wait; on timeout the future is cancelled and a `RuntimeException` is thrown so existing `errors.tolerance` handling routes the failure (Kafka redelivery in at-least-once mode; idempotent retry via `insert_deduplication_token` in exactly-once mode).
```
chernser: There should be general explanation of the issue and how it can be fixed with configuration. Details about `Future` are confusing because it is part of internal implementation.

chernser (changes requested):

gaurav7261:
@chernser please review again

chernser:
Thank you, @gaurav7261 ! 

Will review this week.

chernser (approved):

mzitnik (approved):

PR #761, opened 2026-06-03
Ignore String ".size" subcolumn introduced in ClickHouse 26.1+

### LINKED ISSUE #760: Noisy "Unsupported complex parent type: STRING" warnings on newer ClickHouse versions
### Context

On newer ClickHouse versions, the connector logs a warning for every String column when it describes a table:

```
WARN Unsupported complex parent type: STRING (parent name: 'header.message_id.value')
WARN Unhandled complex type 'STRING' as a child of an array or unexpected subcolumn (parent name: 'borrowers.first_name.encrypted_data', child name: 'borrowers.first_name.encrypted_data.size')
```

### Why

Recent ClickHouse versions added a new `.size` (byte-size) subcolumn to the String type. Unlike array length subcolumns (`.size0`, `.size1`), it has no trailing digit, it's just `.size`.

The connector describes tables with `describe_include_subcolumns=1`, but `Table.java` only filters the numbered form `(SIZE_FIELD_MATCHER = .+\.size[0-9]+$)`. The new `.size` slips through and falls into the default branches of `updateParent()`, logging a warning once per String column:
- scalar String → Table.java:194
- String inside an array → Table.java:132

It's purely version-triggered (appears after a ClickHouse upgrade), unrelated to the user's schema.

### Evidence

Same query, two versions:

| ClickHouse | `DESCRIBE` of `s String` with `describe_include_subcolumns=1` |
|---|---|
| 23.7 | only `s` (String) |
| 26.2.1.390 | `s` (String) **+ `s.size` (UInt64, is_subcolumn=1)** |

CREATE TABLE t (s String) ENGINE = MergeTree ORDER BY tuple();
DESCRIBE TABLE t SETTINGS describe_include_subcolumns = 1;

### Impact

Log noise only. The connector ignores the subcolumn and still builds a correct mapping, so inserts are not affected, but `describeTable` runs periodically, so the warnings repeat.

### Fix

Recognize and skip the no-digit .size String subcolumn in Table.updateParent() (like the existing .size0/.null skips). Add TableTest coverage for String, Array(String), and Tuple(... String) columns carrying a .size subcolumn.

BentsiLeviav:
ClickHouse 26.1+ exposes a ".size" (byte-size) subcolumn on the String type under describe_include_subcolumns. Unlike array length subcolumns (".size0"/".size1"), it has no trailing digit, so Table.updateParent failed to filter it and logged a warning for every String column.

Recognize and skip the ".size" subcolumn when the column (or array element) is a String, matching the existing ".size0"/".null" handling.

closes #760

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Table.java
+++ b/src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Table.java
@@ -22,6 +22,10 @@ import java.util.regex.Pattern;
 public class Table {
     private static final Logger LOGGER = LoggerFactory.getLogger(Table.class);
     private static final Predicate<String> SIZE_FIELD_MATCHER = Pattern.compile(".+\\.size[0-9]+$").asMatchPredicate();
+    // Recent ClickHouse versions (25.x+) expose a ".size" byte-size subcolumn on the String type. Unlike the
+    // array length subcolumns ".size0"/".size1", it has no trailing digit. It is metadata we don't map, so we
+    // skip it instead of warning. See https://github.com/ClickHouse/clickhouse-kafka-connect for context.
+    private static final Predicate<String> STRING_SIZE_FIELD_MATCHER = Pattern.compile(".+\\.size$").asMatchPredicate();
     private static final Pattern MULTIPLE_MAP_VALUES_PATTERN = Pattern.compile("(\\.values)(?=((\\.values)+$))");
 
     private final String name;
@@ -129,6 +133,11 @@ public class Table {
                     case VARIANT:
                         return;
                     default:
+                        if (parentArrayType.getType() == Type.STRING && STRING_SIZE_FIELD_MATCHER.test(childName)) {
+                            // The ".size" byte-size subcolumn of a String element inside an array.
+                            LOGGER.debug("Ignoring String size subcolumn: {}", childName);
+                            return;
+                        }
                         LOGGER.warn("Unhandled complex type '{}' as a child of an array or unexpected subcolumn (parent name: '{}', child name: '{}').",
                                 parentArrayType.getType(), parent.getName(), child.getName() );
                         return;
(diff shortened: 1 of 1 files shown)


mzitnik:
Did we discover it when testing against ClickHouse `head`

BentsiLeviav:
@mzitnik I don't think so, this is just a warning users see - no errors/exceptions/any actual implication to the pipelines.

mzitnik (changes requested):
Lets add  a test to cover the case

src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Table.java
```diff
@@ -129,6 +133,11 @@ private void updateParent(Column parent, Column child) {
                     case VARIANT:
                         return;
                     default:
+                        if (parentArrayType.getType() == Type.STRING && STRING_SIZE_FIELD_MATCHER.test(childName)) {
```
chernser, line 136: This check fixes the issue but not for future added cases. I suggest just commenting out warning message because it was historically added for "just in case" 
Current code tried to update some previously found definitions of Array and Map. Everything else should be ignored. 
Besides we plan to refactor this code and change logic to more streamlined.

src/main/java/com/clickhouse/kafka/connect/sink/db/mapping/Table.java
```diff
@@ -190,6 +199,9 @@ private void updateParent(Column parent, Column child) {
             default:
                 if (child.getName().endsWith(".null")) {
                     LOGGER.debug("Ignoring complex column: {}", child);
+                } else if (parent.getType() == Type.STRING && STRING_SIZE_FIELD_MATCHER.test(child.getName())) {
```
chernser, line 202: This should be in else branch - we need to output in debug what a new column appeared.

chernser (changes requested):

chernser:
Closed as work moved to https://github.com/ClickHouse/clickhouse-kafka-connect/pull/767

PR #786, opened 2026-07-11
Fix source offset tracking after topic-mutating SMTs

### LINKED ISSUE #777: Buffered sink offsets use transformed topic after topic-mutating SMT, causing commits for unassigned partitions
_(CoPilot-assisted bug report)_

### Describe the bug

When `bufferCount > 0`, `ClickHouseSinkTask` tracks flushed offsets using the transformed `SinkRecord.topic()`, `SinkRecord.kafkaPartition()`, and `SinkRecord.kafkaOffset()` values. If a sink Single Message Transform rewrites the record topic, `preCommit()` returns offsets keyed by the transformed/destination topic instead of the original Kafka source topic/partition assigned to the task.

Kafka Connect then cannot commit those offsets because the returned `TopicPartition` keys do not match the task's assigned source partitions. In practice this can look like successful ClickHouse inserts while broker-visible consumer group offsets remain flat or commits are rejected/skipped with warnings about unassigned partitions.

This appears to affect `v1.3.9` and still appears present on `main` as of 2026-07-01. The relevant code path is:

```java
// ClickHouseSinkTask.flushBuffer()
for (SinkRecord record : toFlush) {
    TopicPartition tp = new TopicPartition(record.topic(), record.kafkaPartition());
    long offset = record.kafkaOffset() + 1;
    ...
}
```

The same transformed topic/partition issue appears in `ClickHouseSinkTask.close(...)` when buffered records are removed for revoked partitions:

```java
buffer.removeIf(record -> {
    TopicPartition tp = new TopicPartition(record.topic(), record.kafkaPartition());
    return partitions.contains(tp);
});
```

There is also a related state/original-topic path in `Processing.doLogic(...)`:

```java
String originalTopic = record.getSinkRecord().topic();
```

For sink records that have passed through topic-mutating SMTs, Kafka Connect `SinkRecord` exposes the original Kafka identity via `originalTopic()`, `originalKafkaPartition()`, and `originalKafkaOffset()`. Those values should be used for connector offset tracking and commit ownership. `topic()` may legitimately represent the transformed routing destination after SMTs.

### Steps to reproduce

1. Run `clickhouse-kafka-connect` with internal buffering enabled, for example:

   ```properties
   connector.class=com.clickhouse.kafka.connect.ClickHouseSinkConnector
   topics=source_topic
   exactlyOnce=false
   bufferCount=1000
   bufferFlushTime=10000
   ```

2. Add a sink SMT before the ClickHouse connector receives records that rewrites the topic, for example a route/table SMT like:

   ```java
   public class RouteToTable implements Transformation<SinkRecord> {
       @Override
       public SinkRecord apply(SinkRecord record) {
           return record.newRecord(
               "destination_table",
               record.kafkaPartition(),
               record.keySchema(),
               record.key(),
               record.valueSchema(),
               record.value(),
               record.timestamp(),
               record.headers()
           );
       }

       @Override public ConfigDef config() { return new ConfigDef(); }
       @Override public void configure(Map<String, ?> configs) {}
       @Override public void close() {}
   }
   ```

   Example connector config:

   ```properties
   transforms=route
   transforms.route.type=com.example.RouteToTable
   ```

3. Produce records to `source_topic` and wait for a buffer flush.

4. Observe the offsets returned by `preCommit()` or broker-visible consumer group offsets.

### Expected behaviour

After records are successfully inserted into ClickHouse and the buffer is flushed, `preCommit()` should return commit offsets keyed by the original Kafka source topic/partition, for example:

```text
source_topic-0 -> OffsetAndMetadata(next_source_offset)
```

This is the partition assignment owned by Kafka Connect and the key that can be committed to Kafka.

### Actual behaviour

After the same records are successfully inserted into ClickHouse, `preCommit()` returns commit offsets keyed by the transformed topic name, for example:

```text
destination_table-0 -> OffsetAndMetadata(next_source_offset)
```

That transformed topic is not the Kafka source partition assigned to the task. The framework therefore cannot advance the actual consumer group offsets for `source_topic`, even though the sink may continue inserting data.

### Error log

The exact log wording depends on Kafka Connect version and logging level, but symptoms include warnings/errors equivalent to committing offsets for partitions that are not assigned to the worker/task, for example:

```text
Ignoring/failed commit for unassigned partition destination_table-0
```

Operationally, this can also be observed as:

```text
ClickHouse inserted-record metrics continue increasing
Kafka broker-visible consumer group offsets for the source topic do not advance
Lag grows by approximately the same amount as source latest offsets increase
```

#### Environment

* Kafka-Connect version: observed on a Kafka Connect 3.9.x runtime
* ClickHouse Kafka Connect version: `v1.3.9`; issue appears present on current `main` as of 2026-07-01
* Kafka Connect configuration: `bufferCount > 0`, `exactlyOnce=false`, topic-mutating sink SMT before ClickHouse sink task processes records
* Kafka version: Kafka 3.9.x runtime in the observed environment
* Kafka environment: distributed Kafka Connect worker
* OS: Linux container

#### ClickHouse server

* ClickHouse Server version: not believed to be material to the bug
* ClickHouse Server non-default settings, if any: not material
* `CREATE TABLE` statements for tables involved: any valid destination table matching the transformed topic/table name is sufficient
* Sample data: any valid record that can be inserted after the SMT is sufficient

### Why this matters

Topic-mutating SMTs are a valid Kafka Connect pattern. `SinkRecord.newRecord(...)` preserves the original source topic/partition/offset, and Kafka Connect exposes that original identity specifically so sink connectors can coordinate offset commits correctly after transformations.

The current buffering implementation conflates two different identities:

* `record.topic()` / `record.kafkaPartition()` / `record.kafkaOffset()` - the transformed sink record identity, which may be useful for destination routing
* `record.originalTopic()` / `record.originalKafkaPartition()` / `record.originalKafkaOffset()` - the original Kafka source identity, which should be used for offset commit ownership

For a ClickHouse sink, using the transformed topic to route inserts can be correct, but using it to report Kafka offsets from `preCommit()` is not correct when SMTs mutate the topic.

### Proposed fix

Use the original source Kafka identity for offset tracking and partition ownership whenever available, while keeping the transformed `record.topic()` behavior for ClickHouse table routing.

Conceptually:

```java
private TopicPartition sourceTopicPartition(SinkRecord record) {
    Integer partition = sourceKafkaPartition(record);
    if (partition == null) {
        throw new ConnectException("Cannot determine source Kafka partition for offset commit");
    }
    return new TopicPartition(sourceTopic(record), partition);
}

private String sourceTopic(SinkRecord record) {
    try {
        return record.originalTopic();
    } catch (NoSuchMethodError e) {
        return record.topic();
    }
}

private Integer sourceKafkaPartition(SinkRecord record) {
    try {
        return record.originalKafkaPartition();
    } catch (NoSuchMethodError e) {
        return record.kafkaPartition();
    }
}

private long sourceKafkaOffset(SinkRecord record) {
    try {
        return record.originalKafkaOffset();
    } catch (NoSuchMethodError e) {
        return record.kafkaOffset();
    }
}
```

Then use those helpers in at least:

```java
// ClickHouseSinkTask.flushBuffer()
TopicPartition tp = sourceTopicPartition(record);
long offset = sourceKafkaOffset(record) + 1;
```

```java
// ClickHouseSinkTask.close(...)
TopicPartition tp = sourceTopicPartition(record);
return partitions.contains(tp);
```

And for related state/original-topic reporting:

```java
// Processing.doLogic(...)
String originalTopic = sourceTopic(record.getSinkRecord());
```

If the project still compiles against a Kafka Connect API older than 3.6, the calls can be guarded with `NoSuchMethodError` or reflection. If the project compiles against Kafka Connect 3.6+, direct method calls should work.

### Workaround

The only reliable workaround I know of is to avoid topic-mutating SMTs when `bufferCount > 0`, or to disable connector buffering. That is not practical for high-throughput one-Kafka-topic-to-many-ClickHouse-table routing setups where the transformed topic is used as the ClickHouse table name.

KSHITIZ6341:
Fixes #777.

## Summary

- read original Kafka topic, partition, and offset metadata on Kafka Connect 3.6+ runtimes
- use source identity for buffered offset commits, partition revocation, and inserted-offset reporting
- retain compatibility with older Kafka Connect runtimes by falling back to current record metadata
- bump the connector version to 1.3.12 and add the required changelog entry

The transformed topic remains unchanged for ClickHouse table routing.

## Checklist

- [x] Unit tests covering transformed and legacy-runtime scenarios were added
- [x] A human-readable description of the changes was provided in CHANGELOG

## Verification

- `./gradlew spotlessCheck`
- `./gradlew test --tests 'com.clickhouse.kafka.connect.sink.ClickHouseSinkTaskOffsetTest' --tests 'com.clickhouse.kafka.connect.sink.processing.ProcessingTest'` (18 tests)
- `./gradlew createConfluentArchive`

The Docker-backed feature and integration suites were not run locally; upstream CI can exercise those environments.


chernser:
@KSHITIZ6341 

However I see you are using reflection because we are using old connect-api. Required methods were introduced in 4.0.0 version. So solution would not work with anything older than that (I know, we all should upgrade). 

We need to make a refactoring to separate connect API version, first.

KSHITIZ6341:
Thanks for the clarification. I agree that supporting the newer metadata APIs safely needs a Connect API-version separation first. I’ll keep this PR unchanged rather than extend the reflection-based workaround; the prerequisite refactor should be resolved before revisiting the offset change.

chernser:
I'm closing this PR as stale. 
There is a way to dynamically load metadata extractors depending on the runtime version. 
Or we may be need to separate this code in project and have builds for different runtime versions.

woppa684:
Is addressing this on some roadmap then? We still need to patch the connector every time to make it work.

chernser:
Good day, @woppa684 ! 

It is but not with high priority. I will talk to the team and hope it will be solved in a month. 
Very sorry for this inconvenience.

PR #787, opened 2026-07-11
Limit schema discovery to destination tables

### LINKED ISSUE #632: [schema validation] Reduce the number of DESCRIBE requests
Today, the Connector runs [describe table](https://github.com/ClickHouse/clickhouse-kafka-connect/blob/1d5e2332043ef65263752c73ae978e15fb5e3a95/src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java#L451) to retrieve table metadata and build column mappings. However, the current implementation has a serious flaw: it describes all the tables in a database (see [here](https://github.com/ClickHouse/clickhouse-kafka-connect/blob/1d5e2332043ef65263752c73ae978e15fb5e3a95/src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java#L315)). This could overwhelm the instance with the number of queries, especially if the destination database contains a large number of tables.

We should revisit the logic to run `describe` only on the subset of relevant destination tables. 

might be related to https://github.com/ClickHouse/clickhouse-kafka-connect/issues/184

KSHITIZ6341:
## Summary
- filter table schema discovery before `DESCRIBE TABLE` using the connector's `topics`, `topics.regex`, and `topic2TableMap` configuration
- register runtime-routed destinations lazily so SMTs and database/topic splitting continue to resolve the actual target table
- preserve the existing unfiltered helper overload for compatibility and update schema refreshes to use the requested database
- add database-free selector tests plus Docker-backed assertions that unrelated tables are never described or cached

Fixes #632.

Good catch by @mshustov: the previous mapping pass described every uncached table in the database even when the connector could never write to most of them.

## Compatibility
Connectors with explicit or regex topic subscriptions now eagerly discover only matching destinations. If routing determines a destination at runtime, the writer registers and describes that table on first use. Calls to the existing two-argument `extractTablesMapping` overload retain the previous unfiltered behavior.

## Checks
- `./gradlew spotlessCheck`
- `./gradlew test --tests 'com.clickhouse.kafka.connect.sink.db.DestinationTableFilterTest'` (4 tests passed)
- `./gradlew compileJava compileTestJava`
- `./gradlew createConfluentArchive`

The Docker-backed `ClickHouseHelperClientTest.extractTablesMappingOnlyDescribesRelevantTables`, `ClickHouseWriterTest.configuredTopicsLimitInitialTableMapping`, and `ClickHouseWriterTest.configuredTopicsRegexLimitsInitialTableMapping` checks were added and compiled, but could not run locally because no Docker runtime is available.

## Checklist


chernser:
Good day, @KSHITIZ6341 ! 

Currently the is a problem - new `topic` and `topic.regexp` parameters are get out of sync  with what topics sink is working. Parsing logic of extra table mapping and DB split is another complexity what we have to support. 

Lazy loading of tables is good solution for this problem: 
- logic is simple 
- fix in one place
- already part of the main path. 

Please update the implementation.

chernser (changes requested):
there should be a simpler solution. We can always lazy load tables.

KSHITIZ6341:
Thanks, agreed. I’m simplifying this to start with no eager table mapping, lazily describe each destination on first use, and refresh only cached tables. That avoids duplicating topic/regex, table-map, and DB-split selection logic.

chernser:
closing as stale PR. 
Implementation of this feature should wait some refactoring - today it seems not a good time.

PR #796, opened 2026-07-18
[writing data] RowBinaryInsert with Client V2 

### LINKED ISSUE #795: Writing data with V2 client not optimal
### Description 

V1 write operation worked as true data streaming - data encoded and written to output stream. 
V2 write operation currently writes into buffer and then pass it as input stream to the client. However there is an option to use DataWriter and do streaming in the same way as V1. Moreover it will reduce code duplication because encoding is the same.

chernser:
## Summary
- Rewrite all three insert operations for V2 client using DataWriter 
- Adds script to test connector with multiple different ClickHouse version for local development. 

Closes: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/795
## Checklist


chernser:
Here is benchmark Result After fix 
image
[jmh-results-local-1784352178874.json](https://github.com/user-attachments/files/30205086/jmh-results-local-1784352178874.json)

chernser:
This improvements gives +30% in time increase for big loads (>50k) 
image

benchmark/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ plugins {
 }
 
 val connectorVersion = file("../VERSION").readText().trim()
+val clickHouseDriverVersion = "0.9.5"
```
mzitnik: Should we use the same version as the connector?

mzitnik (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+# 1.4.1, 2026-07-17
+
+## Improvements
+* RowBinary inserts with client V2 now stream data directly to the network output stream via the client's
```
chernser, line 4: Add documentation for this and configuration of client that may affect network

benchmark/build.gradle.kts
chernser: yes. Will fix.

chernser: done.

CHANGELOG.md
chernser: updating doc.

Issue #726, opened 2026-04-17
Support incompatible Avro data types


kurnoolsaketh:
Fix the connector to support the incompatible schema types identified in https://github.com/ClickHouse/clickhouse-kafka-connect/issues/635:

- [ ] `decimal` logical type (see: `src/testFixtures/avro/schemas/incompatible/decimal_fixed_logical.json`)
- [ ] nullable unions (see: `src/testFixtures/avro/schemas/incompatible/union_null_string_int.json`)
- [ ] record unions (see: `src/testFixtures/avro/schemas/incompatible/union_two_records.json`)

This list is evolving as we identify more incompatible schemas.

29shivam:
Hi @chernser — I'd like to pick this up it's close to the data-type / writer work I did on #729.

Here's the logic I'd propose, working through the list:

- Nullable unions(`["null", "string"]`) — standard Avro nullability. Once the Avro converter turns this into a Connect optional field, the connector should map it straight to ClickHouse `Nullable(T)` on the inner type. The write side already handles `Nullable`, so this is mostly unwrapping the union to its non-null branch in the schema/convert layer and carrying the optionality through. Highest-value and least ambiguous, so I'd start here.

- `decimal` logical type (`decimal_fixed_logical.json`) — map the Avro `decimal` logical type (Connect `Decimal`, which carries precision/scale) to ClickHouse `Decimal(P, S)`, converting the bytes/fixed representation to `BigDecimal`. The writer already serializes `Decimal` columns via `BigDecimal`, so the gap is on the convert side — reading the logical type + scale and producing the `BigDecimal`.

- Record unions (`union_two_records.json`) — the genuinely ambiguous one: a union of two distinct record types has no single obvious ClickHouse target. I'd want your call before implementing — map to `Variant`, or reject with a clear non-retryable `DataException` naming the field? Happy either way.

I'd drive it all with the existing `src/testFixtures/avro/schemas/incompatible/*` fixtures.

Does that direction and ordering (nullable unions + decimal first, record unions once we align) look right? I'll hold off on a PR until you confirm.

chernser:
@29shivam 

Yes, please go ahead! 
I've created three sub-issues to track implementation separately (3 PRs expected)> 

`Variant` is for simple types like `String` and `Int32`. Best match I see is `JSON`. 
`Dynamic` would also match but we need to plan bigger support for it.

PR #803, opened 2026-07-31
[avro] Support nullable multi-type unions as Variant (#799)

### LINKED ISSUE #799: [avro] Nullable unions
### Description 

nullable unions (see: src/testFixtures/avro/schemas/incompatible/union_null_string_int.json)

29shivam:
## Summary

Fixes #799 (part of #726).

Confluent's Avro converter turns a multi-branch union like `[null, string, int]` into an **optional** Connect union struct (`io.confluent.connect.avro.Union`) with one optional field per non-null branch. The connector already handles this end to end:
- `Column.connectTypeToClickHouseType` maps the union struct → `Variant(String, Int32)`.
- The RowBinary Variant write path picks the non-null branch, resolves its global discriminator (case-insensitively matching the Connect type to the ClickHouse variant type), and writes discriminator + value — or the ClickHouse null discriminator (`255`) when no branch is set.

Per @chernser's guidance, `Variant` is the right target here since the branches are simple types (`String`, `Int32`).

Changes:
- **Add `AvroNullableUnionVariantTest`** — drives the string branch, the int branch, and the all-null case through the convert (`StructToJsonMap`) + Variant write path, asserting the discriminator + value bytes. Runs without a live ClickHouse.
- **Promote `union_null_string_int.json`** from `incompatible/` to `compatible/` so the Avro integration test exercises it end to end.
- CHANGELOG entry under `1.4.1 (unreleased)`.

**Verification:** feature test passes locally; the full end-to-end path runs via the integration test in CI.

## Checklist


29shivam:
@chernser ready for review — nullable multi-type unions (#799). Per your note, `Variant` fits here since the branches are simple types (`String`, `Int32`). Feature test drives the string/int/all-null cases through the Variant write path; fixture promoted to `compatible`.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+# 1.4.1, (unreleased)
```
chernser: please update to latest one

chernser (changes requested):
Is there code changes? I see only test. 
tests should cover more unions - because not all might fit Variant.

29shivam:
Thanks @chernser!

- **Code vs test:** you're right — there's no new *production* code here. The union→Variant path already exists (`connectTypeToClickHouseType`/`resolveUnionType` map the Confluent union struct to `Variant(...)`, and the RowBinary Variant write path resolves the per-value discriminator), so this fixture was already handled correctly. The PR's value is promoting it from `incompatible/` → `compatible/` so the integration test exercises it, plus locking in coverage. If you'd rather not have a test-only PR, I'm happy to fold this into another PR or close it — your call.

- **More union coverage:** expanded the test — it now covers `Variant(String, Int32)` (string / int / all-null), `Variant(String, Int32, Bool)` (bool branch + discriminator), and a suspicious same-group numeric union `[int, long]` that falls back to `Nullable(String)` rather than a Variant — so 'not everything fits a Variant' is covered.

- **CHANGELOG:** rebased onto latest `main`; the entry now sits under `1.5.0 (not published)`.

29shivam:
The one red CI cell (`one_shard_three_replicas 26.3 V1`) is a runner/infra failure, not this change — it died at container startup with `Can't get Docker image: testcontainers/ryuk:0.14.0` and the subsequent `docker compose up` failing, so every container-based test errored at init (all `initializationError`, no assertions). The other 32 cells are green, including the cluster/standalone cells that run the Avro integration test with the promoted `union_null_string_int` fixture — so the fixture passes end to end. Could you re-run that one job when you get a chance? 🙏

chernser (approved):

29shivam:
Thank you @chernser.

PR #814, opened 2026-08-05
[avro] Support record unions as JSON (#800)

### LINKED ISSUE #800: [Avro] Record Union
### Description 
record unions (see: src/testFixtures/avro/schemas/incompatible/union_two_records.json)

Avro allows to have: 
```
{"name": "payload", "type": [{
        "type": "record",
        "name": "TypeA",
        "fields": [
          {"name": "label", "type": "string"}
        ]
      }, {
        "type": "record",
        "name": "TypeB",
        "fields": [
          {"name": "count", "type": "int"}
        ]
      }]
``` 

The only matching type in ClickHouse is JSON because it has no restrictions on value schema. 
Dynamic and Variant is not a match for complex objects.

29shivam:
## Summary

Fixes #800 (part of #726). Implements the direction from [your comment](https://github.com/ClickHouse/clickhouse-kafka-connect/issues/800#issuecomment-5186529775) @chernser.

Confluent's Avro converter turns a union of two records (`[TypeA, TypeB]`) into a Connect union struct (`io.confluent.connect.avro.Union`) keyed by branch record name. Per your **Item 1**, the value is kept in this **tagged form** (which branch it was is preserved) — no serialization change, a ClickHouse `JSON` column stores it as-is.

Per your **Item 2**, writing to a `JSON` column in RowBinary needs `input_format_binary_read_json_as_string=1`, so the Avro integration harness now routes JSON-target fixtures through a dedicated connector config.

Changes:
- **`clickhouse_sink_avro_json.json`** + `SinkConfigs.AVRO_JSON` + **`setupAvroConnectorWithJson`** — Avro converter with the JSON-as-string setting. `avroSchemaTest` routes to it when a fixture's target table has a `JSON` column.
- **`AvroRecordUnionJsonTest`** — asserts the tagged-by-branch JSON serialization of a record union (parses the output; no live ClickHouse needed).
- **Retarget `union_two_records`** from `Variant(Tuple, Tuple)` → `JSON` and promote it from `incompatible/` → `compatible/`.
- CHANGELOG under `1.5.0`.

**On the two branches you mentioned:** this covers the `input_format_binary_read_json_as_string=1` (happy) path end to end. For the `=0` + JSON-present branch — today the connector throws `Writing JSON in binary is not supported yet…` I'd like your call on whether that should stay a task failure or route to DLQ before I add that coverage (happy to do it in this PR or a follow-up).

**Verification:** feature test passes locally the end-to-end Avro→JSON path runs via the integration test in CI.

## Checklist


29shivam:
@chernser ready for review — record union → JSON (#800), implementing the direction from your comment on the issue: the value is kept in tagged form, and JSON-target fixtures are routed through a new `setupAvroConnectorWithJson` (Avro converter + `input_format_binary_read_json_as_string=1`). CI is green (33/33), including the integration cells that run the retargeted `union_two_records` fixture end to end.

One open question I flagged in the description rather than guess: for the `input_format_binary_read_json_as_string=0` + JSON-present branch, should the connector keep failing the task (as today) or route to DLQ? Happy to add that coverage here or as a follow-up once you decide.

29shivam:
@chernser — gentle bump here too whenever you have a moment. Green and mergeable (record-union → JSON, #800). Still happy to take your call on the `input_format_binary_read_json_as_string=0` branch question. No rush.

src/test/java/com/clickhouse/kafka/connect/sink/db/AvroRecordUnionJsonTest.java
```diff
@@ -0,0 +1,83 @@
...
+ * Verifies how a record union {@code [TypeA, TypeB]} serializes to JSON for a ClickHouse {@code JSON}
+ * column (issue #800, from #726). Confluent's Avro converter turns a union of two records into a
+ * Connect struct named {@code io.confluent.connect.avro.Union} with one optional struct field per
+ * branch (keyed by the branch record name). Per the maintainer, the value is kept in this tagged
+ * form (which branch it was is preserved); ClickHouse's {@code JSON} type stores it as-is.
+ *
+ * <p>Asserted by parsing the output rather than string-matching, since field order in the tag object
+ * is not significant.
+ */
+public class AvroRecordUnionJsonTest {
```
chernser, line 27: This test should interact with database in test container.

chernser (changes requested):

chernser:
@29shivam 
Sorry - I've missed the question about branches. We should still fail on JSON as binary because it is not supported and is configuration issue. 
Current main test is in main source and it just need to be against real ClickHouse instance - that way we verify data is  processed.

29shivam:
Thanks @chernser, that makes sense. Pushed an update:

- Dropped the in-memory unit test and moved the check into the Avro integration harness, so it now runs against real ClickHouse and reads the rows back to confirm the tagged-by-branch JSON actually landed (not just the row count).
- Good call on the `input_format_binary_read_json_as_string=0` branch too, kept it failing the task since it's a config issue, so no change needed there.

One thing the end-to-end run surfaced: ClickHouse stores the simple record name in the JSON column (`{"TypeA": {...}}`), while the Avro input still needs the fully-qualified branch name, so the fixture carries both forms on purpose. Ready for another look whenever you have a moment.

chernser (approved):

29shivam:
Thanks for the review and merge @chernser! Really enjoyed working through this one, looking forward to contributing more.

Issue #587, opened 2025-09-11
Handling ClickhouseExceptions as RetriableExceptions


Vaishnavi-Arisetty:
### Describe the bug
The ClickHouse Kafka Sink Connector should have the capability to retry when a ClickHouseException occurs — for example, in cases such as memory limit exceeded errors from ClickHouse or temporary network unavailability.

At the moment, these exceptions are neither retried nor routed to the configured DLQ topic. Instead, the connector task stays alive but the affected records are effectively stuck and never delivered. This makes it difficult to ensure reliable and fault-tolerant message processing.

### Steps to reproduce

1. Start Kafka, Kafka Connect, and ClickHouse.
2. Configure ClickHouse with a very low max_memory_usage (e.g. 10MB) to trigger memory errors.
3. Deploy the ClickHouse Kafka Sink Connector with: 

        "connector.class": "com.clickhouse.kafka.connect.ClickHouseSinkConnector",
        "tasks.max": "1",
        "transforms": "dropPrefix",
        "transforms.dropPrefix.regex": "^[^.]+\\.(.*)",
        "ssl": "false",
        "database": "${dbname}",
        "hostname": "${host}",
        "password": "${password}",
        "consumer.metadata.max.age.ms": "1000",
        "consumer.override.max.poll.records": "20000",
        "errors.deadletterqueue.topic.replication.factor": "1",
        "value.converter": "org.apache.kafka.connect.json.JsonConverter",
        "errors.retry.timeout": "60",
        "transforms.dropPrefix.replacement": "$1",
        "table.name.format": "$${topic}",
        "topics.regex": "^${databaseName}\\..*",
        "transforms.dropPrefix.type": "org.apache.kafka.connect.transforms.RegexRouter",
        "port": "8123",
        "errors.deadletterqueue.topic.name": "dlq",
        "value.converter.schemas.enable": "false",
        "consumer.override.max.partition.fetch.bytes": "5242880",
        "errors.tolerance": "all",
        "exactlyOnce": "false",
        "username": "${userName}"

4. Create a ClickHouse table and a Kafka topic (test-clickhouse-topic).
5. Produce a large message (e.g. ~50MB payload) to the topic.
6. Observe in the connector logs that ClickHouse throws Memory limit exceeded (or stop the ClickHouse container to simulate network unavailability).
7. Check connector status (/connectors/<name>/status) → task remains RUNNING.
[...]

Issue #455, opened 2024-10-03
Add more customization to the retry mechanisms


Paultagoras:
- Allow for limited retries for exceptions (limited by exception code and/or all retriable exceptions)
- Allow customizing the retriable exceptions list

PR #817, opened 2026-08-27
Add retryOnSocketException to retry SocketException in handleException

billykern:
## Summary
A `java.net.SocketException` (broken pipe, connection reset) from the ClickHouse client is currently non-retriable in `Utils.handleException`: it isn't a `SocketTimeoutException` and its message doesn't start with "Read/Write timed out after", so with `errors.tolerance=none` the task is killed, and with `errors.tolerance=all` the batch is dropped.

This adds a `retryOnSocketException` setting (default `false`, preserving current behavior). When enabled, `SocketException` is treated like the existing timeout cases and throws `RetriableException`, letting the Connect framework redeliver the batch with backoff. It matters more with `client_version=V2`: the pooled, persistent connections mean a single server-side connection reset (replica restart, upgrade, scale event) surfaces as a broken pipe on every task's next insert at once. Observed with v1.5.0 on ClickHouse Cloud: 55 tasks across two connectors failed within seconds on

```
com.clickhouse.client.api.DataTransferException: Insert failed (Attempt: 1/4 ...)
Caused by: java.net.SocketException: Broken pipe
    at ...ChunkedOutputStream.write
    at ...ClickHouseWriter.writeRecordsJson
```

The client's own retry loop does not cover this fault (`HttpAPIClientHelper.shouldRetry` only matches no-response, connect/pool timeouts and server-retryable codes), so the connector is the right place for it.

Retry safety: a broken pipe during the request body means the server never received a complete insert, so nothing was applied. A reset while reading the response has the same ambiguity as the already-retried read timeout and code 319 (`UNKNOWN_STATUS_OF_INSERT`), covered by the same `insert_deduplication_token` the connector sets on every insert.

Related: #455 and #587 ask for a configurable retriable list; this adds one opt-in case and doesn't preclude that.

## Checklist

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/util/Utils.java
+++ b/src/main/java/com/clickhouse/kafka/connect/util/Utils.java
@@ -69,7 +69,7 @@ public class Utils {
      * @param e Exception to check
      */
 
-    public static void handleException(Exception e, boolean errorsTolerance, Collection<SinkRecord> records) {
+    public static void handleException(Exception e, boolean errorsTolerance, boolean retryOnSocketException, Collection<SinkRecord> records) {
         LOGGER.warn("Deciding how to handle exception: {}", e.getLocalizedMessage());
 
         //Let's check if we have a ClickHouseException to reference the error code
@@ -116,7 +116,7 @@ public class Utils {
         } else if (rootCause instanceof UnknownHostException) {
             LOGGER.warn("UnknownHostException thrown, wrapping exception: {}", e.getLocalizedMessage());
             throw new RetriableException(e);
-        } else if (rootCause instanceof SocketException) {
+        } else if (retryOnSocketException && rootCause instanceof SocketException) {
             LOGGER.warn("SocketException thrown, wrapping exception: {}", e.getLocalizedMessage());
             throw new RetriableException(e);
         } else if (rootCause instanceof IOException) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/clickhouse/kafka/connect/util/Utils.java
```diff
@@ -108,6 +109,9 @@ public static void handleException(Exception e, boolean errorsTolerance, Collect
         } else if (rootCause instanceof UnknownHostException) {
             LOGGER.warn("UnknownHostException thrown, wrapping exception: {}", e.getLocalizedMessage());
             throw new RetriableException(e);
+        } else if (rootCause instanceof SocketException) {
+            LOGGER.warn("SocketException thrown, wrapping exception: {}", e.getLocalizedMessage());
```
chernser, line 120: All good but this will change behavior. please add a configuration flag to enable this handling. by default flag should be `off` to preserve old behavior. 
Thanks!

chernser (changes requested):

src/main/java/com/clickhouse/kafka/connect/util/Utils.java
billykern: Good point, I didn't even think about that, updating!

chernser (approved):

PR #819, opened 2026-08-27
Apply the retriable error-code list to client V2 ServerException

### LINKED ISSUE #818: handleException error-code retry list never matches client V2 ServerException
`Utils.handleException` checks `rootCause instanceof com.clickhouse.client.ClickHouseException` to decide whether a server error code (3, 107, 159, 164, 202, 203, 209, 210, 241, 242, 252, 285, 319, 425, 999) is retriable. With `client_version=V2` the server error arrives as `com.clickhouse.client.api.ServerException`, which extends `com.clickhouse.client.api.ClickHouseException`, a different class, so the check never matches and none of those codes are retried at the connector level. `getRootCause(e, true)` also never stops on it, so the switch sees the deepest cause instead.

Reproduced on v1.5.0: `Utils.handleException(new RuntimeException(new ServerException(252, "Too many parts", 500)), false, records)` throws `RuntimeException` (task failure) rather than `RetriableException`.

client-v2's own `ServerRetryable` loop masks this for up to `client_max_retries` immediate attempts (default 3, tunable via `jdbcConnectionProperties=retry=N`), after which the task fails with `errors.tolerance=none` or the batch is dropped with `errors.tolerance=all`, instead of backing off via `RetriableException` as V1 does. Affects e.g. `TOO_MANY_PARTS` (252) and `MEMORY_LIMIT_EXCEEDED` (241) under sustained pressure, where a few immediate retries cannot help.

Suggested fix: match `ServerException` (using `getCode()`) alongside the V1 exception in both `getRootCause` and `handleException`, or delegate to `ServerException.isRetryable()`, which carries the same code list.

Related: #587 (possibly the same symptom on V2).

billykern:
## Summary
Fixes #818.

`Utils.handleException` decides whether a ClickHouse server error code is retriable by checking `rootCause instanceof com.clickhouse.client.ClickHouseException`, the client V1 exception. With `client_version=V2` the server error arrives as `com.clickhouse.client.api.ServerException`, which extends a different `ClickHouseException` (`com.clickhouse.client.api`), so the check never matches: `getRootCause(e, true)` walks past it to the deepest cause and the error-code switch is skipped. The result is that every code in the retriable list (`TOO_MANY_PARTS`, `MEMORY_LIMIT_EXCEEDED`, `KEEPER_EXCEPTION`, ...) fails the task with `errors.tolerance=none`, or drops the batch with `errors.tolerance=all`, once client-v2's few immediate in-process attempts are exhausted. V1 throws `RetriableException` and lets the Connect framework back off.

This makes `getRootCause` stop on either exception type and normalizes the code (`getErrorCode()` for V1, `getCode()` for V2) before the existing switch, so the connector keeps a single retriable list and V1 and V2 behave identically. No change for V1.

Tests cover root-cause resolution, a retriable code (252) and a non-retriable code (60) through `ServerException`.

## Checklist

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/util/Utils.java
+++ b/src/main/java/com/clickhouse/kafka/connect/util/Utils.java
@@ -1,6 +1,7 @@
 package com.clickhouse.kafka.connect.util;
 
 import com.clickhouse.client.ClickHouseException;
+import com.clickhouse.client.api.ServerException;
 import com.clickhouse.kafka.connect.sink.data.Record;
 import com.clickhouse.kafka.connect.sink.db.mapping.Column;
 import com.clickhouse.kafka.connect.sink.db.mapping.Table;
@@ -37,10 +38,10 @@ public class Utils {
     }
 
     /**
-     * This will drill down to the first ClickHouseException in the exception chain
+     * This will drill down to the first ClickHouse server exception (client V1 or V2) in the exception chain
      *
      * @param e Exception to drill down
-     * @return ClickHouseException or null if none found
+     * @return the server exception, otherwise the deepest cause, or null if none found
      */
     public static Exception getRootCause(Exception e, Boolean prioritizeClickHouseException) {
         if (e == null)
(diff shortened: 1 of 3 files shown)


src/main/java/com/clickhouse/kafka/connect/util/Utils.java
```diff
@@ -37,25 +38,33 @@ public static Exception getRootCause(Exception e) {
...
 
         return runningException instanceof Exception ? (Exception) runningException : null;
     }
 
+    private static boolean isServerError(Throwable t) {
+        return t instanceof ClickHouseException || t instanceof ServerException;
+    }
+
+    private static int getServerErrorCode(Throwable t) {
+        return t instanceof ServerException ? ((ServerException) t).getCode() : ((ClickHouseException) t).getErrorCode();
```
chernser: this should have check from `isServerError` to be safe.  So I suggest moving body of `getServerErrorCode` to the method it uses and branch where `isServerError(t) == true`.

chernser (changes requested):

chernser:

src/main/java/com/clickhouse/kafka/connect/util/Utils.java
billykern: thanks for catching, updating

chernser (approved):

PR #825, opened 2026-09-01
[Data] Conversion of number types to wider ones. 

### LINKED ISSUE #642: ClassCastException when using type widening/promotion
### Describe the bug
First of all: I'm not sure this is a bug report or a feature request. The problem we're having at the moment is that we have multiple versions of an Avro schema, in which some type promotion has occurred. In our case, in version 1 of the schema, some field was an `int` but in version 2 the field was promoted to a `long`. Also the column in the database was updated from an `Int32` to an `Int64`.

So far so good. Unfortunately, because of some other constraints, we cannot decode with the latest schema but we have to decode with the exact schema the message was produced with. This means that in this example case, the value arrives in the sink connector as an `Integer` and the column expects a `Long`. This results in 
```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long
```
in:
https://github.com/ClickHouse/clickhouse-kafka-connect/blob/d9e5aee46bc0084a9f62436ea2b41884526c28c0/src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java#L680-L680

### Expected behaviour
We previously set `bypassRowBinary` to `true`, so the value is converted to JSON first to make it work. We would like to get rid of that, so for now we have our own fork of this connector, where we implemented this specific cast as 
```java
 BinaryStreamUtils.writeInt64(stream, ((Number) value).longValue());
```

We were wondering whether type promotion/widening could be a bit more supported, maybe even behind some flag? The typical use cases (for [Avro](https://avro.apache.org/docs/1.12.0/specification/)) are:
- `int` to `float`, `double`  or `long`
- `long` to `float` or `double`
- `float` to `double`
- `string` to `bytes`
- `bytes` to `string`

For us, at least the numeric cases should be supported.

chernser:
## Summary
- Updates schema validation to allow only smaller numbers be written to same size or bigger columns 
- Cast number values to Number  and take needed part. Resolves class cast exception 

Closes: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/642

## Checklist

### DIFF
--- a/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
+++ b/src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
@@ -1875,32 +1875,45 @@ public class ClickHouseSinkTaskWithSchemaTest extends ClickHouseBase {
         ClickHouseTestHelpers.dropTable(chc, topic);
     }
 
-    @Test
-    public void testInt32ToInt64ConversionWithConnectSchema() throws Exception {
-        String topic = createTopicName("test_int32_to_int64_connect");
+    private static Stream<Arguments> integerTypesConnectSchemaProvider() {
+        List<Arguments> args = new ArrayList<>();
+        List<String> targetTypes = List.of("Int64", "Int128");
+        for (String targetType : targetTypes) {
+            args.add(Arguments.of("INT8_to_" + targetType, Schema.INT8_SCHEMA, (byte) 12, (byte) 34, Arrays.asList((byte) 100, (byte) 110, (byte) 120), targetType));
+            args.add(Arguments.of("INT16_to_" + targetType, Schema.INT16_SCHEMA, (short) 1234, (short) 5678, Arrays.asList((short) 100, (short) 200, (short) 300), targetType));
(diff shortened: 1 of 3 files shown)


src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
```diff
@@ -1871,6 +1875,178 @@ public void testAvroDateAndTimeTypes() throws Exception {
         ClickHouseTestHelpers.dropTable(chc, topic);
     }
 
+    @Test
```
mzitnik: Do you also want to add 8-> 16, 32, 64, 128, 256, and other different permutations

mzitnik (approved):
Approved with note

src/test/java/com/clickhouse/kafka/connect/sink/ClickHouseSinkTaskWithSchemaTest.java
chernser: yes, it will not hurt.

chernser: done.

Issue #712, opened 2026-04-08
Update design doc to be more agent friendly


kurnoolsaketh:
Currently, the design doc embeds diagrams as PNG's that are difficult for agents to parse and understand. Convert the diagrams to text so that we can reference it explicitly from AGENTS.md.

A first stab at this was proposed here: https://github.com/ClickHouse/clickhouse-kafka-connect/pull/703 - we can use this as a starting point.

AzazelSensei:
Opened https://github.com/ClickHouse/clickhouse-kafka-connect/pull/826 — ASCII versions of the design diagrams, with AGENTS.md pointing at them.

PR #826, opened 2026-09-02
docs: convert DESIGN.md diagrams to text #712

AzazelSensei:
## Summary
Replace the PNG embeds in docs/DESIGN.md with ASCII diagrams and link them from AGENTS.md. #712

The original images stay in docs/ as archive links next to each diagram.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,6 +18,14 @@ Key design goals:
 
 Full design rationale: [`docs/DESIGN.md`](./docs/DESIGN.md)
 
+The diagrams in that doc are ASCII so they can be cited directly:
+- [Deduplication flow](./docs/DESIGN.md#deduplication-flow)
+- [High-level architecture](./docs/DESIGN.md#high-level-architecture)
+- [Initial `NONE` state](./docs/DESIGN.md#initial-state-none)
+- [Full state machine](./docs/DESIGN.md#full-state-machine)
+- [Cropping batches](./docs/DESIGN.md#cropping-batches)
+- [Invalid states](./docs/DESIGN.md#invalid-states)
+
 ### Package Structure
 
 All source code lives under `src/main/java` in the `com.clickhouse.kafka.connect` package:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,14 +18,6 @@ Key design goals:
 
 Full design rationale: [`docs/DESIGN.md`](./docs/DESIGN.md)
 
-The diagrams in that doc are ASCII so they can be cited directly:
-- [Deduplication flow](./docs/DESIGN.md#deduplication-flow)
-- [High-level architecture](./docs/DESIGN.md#high-level-architecture)
-- [Initial `NONE` state](./docs/DESIGN.md#initial-state-none)
-- [Full state machine](./docs/DESIGN.md#full-state-machine)
-- [Cropping batches](./docs/DESIGN.md#cropping-batches)
-- [Invalid states](./docs/DESIGN.md#invalid-states)
-
 ### Package Structure
 
 All source code lives under `src/main/java` in the `com.clickhouse.kafka.connect` package:
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -36,7 +36,27 @@ We dismissed solutions using a two-phase commit or storing offsets in ClickHouse
 
 The current solution exploits a property of ClickHouse, which can initially cause some [getting-started confusion](https://clickhouse.com/blog/common-getting-started-issues-with-clickhouse). If identical inserts are made to ClickHouse within a certain time period, they are de-duplicated. This is known as the [replicated_deduplication_window](https://clickhouse.com/docs/en/operations/settings/merge-tree-settings/#replicated-deduplication-window) and is enabled by default on replicated instances. When data is inserted into ClickHouse, it creates one or more blocks (parts). In replicated environments, such as ClickHouse Cloud, a hash is written in ClickHouse Keeper. Subsequent inserted blocks are compared against these hashes and ignored if a match is present. This is useful since it allows clients to safely retry inserts in case of no acknowledgment from ClickHouse, e.g., because of a network interruption. This requires blocks to be identical, i.e., the same size with the same rows in the same order. These hashes are stored for only the most recent 100 blocks, although this [can be modified](https://clickhouse.com/docs/en/operations/settings/merge-tree-settings/#replicated-deduplication-window). Note higher values will slow down inserts due to the need for more comparisons.
 
-![deduplication](./deduplication.png)
+#### Deduplication flow
+
+```
+Insert query
+    │
+    ▼
+ClickHouse: create data block(s)
+    │  per block
+    ▼
+Calculate hash sum for data block
+    │
+    ▼
+Check if hash sum already exists  ──read──►  Keeper deduplication log
+    │
+    ├─ yes ──► Ignore data block
+    │
+    └─ no  ──► Create data part on storage
+               + write hash sum to Keeper
+```
+
+Original diagram: [deduplication.png](./deduplication.png)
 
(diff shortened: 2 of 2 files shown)


chernser:
@AzazelSensei 

docs/DESIGN.md
```diff
@@ -36,7 +36,27 @@ We dismissed solutions using a two-phase commit or storing offsets in ClickHouse
...
 
-![deduplication](./deduplication.png)
+#### Deduplication flow
+
+```
+Insert query
```
chernser: Image visually was more readable by humans. For AI Agents this is the problem. 
This diagram would be a problem for LLMs to because they hard to understand 2D space. 

Please make it a text sequence diagram - so it is readable and editable by humans
And please add preudo code underneath for LLMs (it it can be just list of steps).

docs/DESIGN.md
```diff
@@ -49,7 +69,31 @@ If we could ensure these properties hold true, we could exploit ClickHouse’s a
...
+                           create batch per Topic+Partition
+                                    │  batch
+                                    ▼
+                           state machine (single worker thread)
+                             1. look up state of previous insert
+                             2. generate consistent batches
+                             3. (re-) insert batches
+                             4. store state of current insert
+                                    │
+                    ┌───────────────┴────────────────┐
```
chernser: please do the same as for first one. 
it is also not clear why there is a for here - on sequence diagram it would be much clear.

docs/DESIGN.md
```diff
@@ -49,7 +69,31 @@ If we could ensure these properties hold true, we could exploit ClickHouse’s a
 
 To ensure we always formulate identical batches to ClickHouse, we need to maintain a state of our current processing and a record of the previous insert. We can use this information to slice and merge batches to ensure the consistency of inserts. We do this using a state machine per topic and partition to align with how processing can be distributed in Kafka and how offsets are tracked. Specifically, for each `put` of records, we generate a batch per topic and partition, look up the previous state for this topic/partition and generate consistent inserts to ClickHouse based on a set of rules implemented as a state machine. This can be summarized as shown below:
 
-![architecture](./architecture.png)
+#### High-level architecture
+
+```
+Kafka  -- 'put' of records -->  Kafka Connect Sink
```
chernser: I remember there was a good big detailed image. Looks like something is missing.

docs/DESIGN.md
AzazelSensei: Replaced that 2D ASCII with a mermaid sequence diagram and a step list under it. Original PNG is still linked.

docs/DESIGN.md
```diff
@@ -72,7 +116,21 @@ For each topic/partition batch, we record the following information:
 
 When handling a batch for a specific topic and partition, we look up this information for the previous insert. If the latest state is `AFTER,` we know the previous insert was successful. Conversely, `BEFORE` means the connector could not confirm whether ClickHouse received and acknowledged the data. The absence of either state means either this is the first time we have seen data for this topic or partition, or we weren't able to set even set a `BEFORE` state. These are logically identical. This is the simplest case in our state machine:
 
-![start state](./start_state.png)
+#### Initial state (`NONE`)
+
+```
+Stored state: NONE
```
chernser: Image was read better - replace with table. This is most simple state machine description.

AGENTS.md
```diff
@@ -18,6 +18,14 @@ Key design goals:
 
 Full design rationale: [`docs/DESIGN.md`](./docs/DESIGN.md)
 
+The diagrams in that doc are ASCII so they can be cited directly:
```
chernser: There is already link to DESIGN.md. 
Please revert this change. 

TOC should be in DESIGN.md - because it is how it will be updated. Here links will break.

docs/DESIGN.md
```diff
@@ -85,17 +143,56 @@ For subsequent inserts, we should have a previous state for each topic/partition
...
 
 There is one case not covered by the above. We may simply get data for a partition and batch whose `minOffset` is less than the `minOffset` of the previous state. This is shown below:
 
-![cropping batches](./cropping_batches.png)
+#### Cropping batches
+
+When the incoming batch starts below the stored `minOffset`, the prefix is dropped and the remainder is evaluated with the rules above:
+
+```
+previous:          [storedMin ──────── storedMax]
```
chernser: please use examples and may be standard math notation for better reading. 
`^^^^^` is not clear what it is.

chernser (changes requested):

docs/DESIGN.md
AzazelSensei: Switched this to a mermaid sequence plus a step list, same as the first diagram. The loop is each topic+partition on a single worker thread.

docs/DESIGN.md
AzazelSensei: Filled in the Keeper vs ClickHouse split and the per-topic+partition loop from the original image.

docs/DESIGN.md
AzazelSensei: Replaced this with a table.

AGENTS.md
AzazelSensei: Reverted. TOC is in DESIGN.md now.

docs/DESIGN.md
AzazelSensei: Dropped the carets. Example is stored [10, 19], current [5, 25]; drop offset < 10, remainder [10, 25].

docs/DESIGN.md
```diff
@@ -49,7 +96,38 @@ If we could ensure these properties hold true, we could exploit ClickHouse’s a
...
+    participant CH as ClickHouse
+
+    Kafka->>Sink: put(records)
+    Sink->>Sink: create a batch per topic+partition
+    loop each topic+partition (single worker thread)
+        Sink->>SM: batch
+        SM->>Keeper: look up previous insert state
+        Keeper-->>SM: flag + [storedMin, storedMax]
+        SM->>SM: generate consistent batches
+        SM->>CH: (re-)insert batches
```
chernser, line 117: there should be `SM->>Keeper: Store BEFORE_PROCESSING state`
after insert :  `SM->>Keeper: Store AFTER_PROCESSING state` 

These are critical steps protecting helping to recover from failure.

AzazelSensei: Added those two steps. The sequence now stores BEFORE_PROCESSING in Keeper before the insert and AFTER_PROCESSING after it, with the same order in the step list.

docs/DESIGN.md
```diff
@@ -85,17 +174,59 @@ For subsequent inserts, we should have a previous state for each topic/partition
 
 This is summarized below:
 
-![full state machine](./full_state_machine.png)
+#### Full state machine
+
+Comparison of previous insert `[storedMin, storedMax]` with current batch `[currentMin, currentMax]`.
```
chernser, line 181: Please consider this table if more readable: 
```
Stored State       | Overlaping State | Action 
-------------------+------------------+-------------
 NONE              | Any              | Insert
 BEFORE_PROCESSING | Zero             | Insert 
 BEFORE_PROCESSING | New or Same      | Insert dedup
 BEFORE_PROCESSING | Contains         | ERROR 
 BEFORE_PROCESSING | Overlapping      | Insert stored + Insert new
 AFTER_PROCESSING  | Same, Contains   | Ignore
 AFTER_PROCESSING  | Zero             | Insert
 AFTER_PROCESSING  | New              | Insert 
 AFTER_PROCESSING  | Overlapping      | Insert only new records
 AFTER_PROCESSING  | Contains         | Error or ignored if `tolerateStateMismatch=true` 
  ```

May be add as addition.

chernser (approved):

docs/DESIGN.md
AzazelSensei: Added that table under Full state machine as a compact view of the same rules.

PR #828, opened 2026-09-04
[schema] Sticky Replica Routing on Schema related error. 

### LINKED ISSUE #827: Error handling: Use Cloud feature to send schema and insert requests to same replica
### Description 

Precondition: 
- Sink used with ClickHouse Cloud
- Table schema changed 

There is a problem - when table schema is detected on Sink it is usually an error. Today ClickHouse Sink detects this error and does two things: 
- refreshes schema 
- retries insert 
Both this operations can land on different replicas and error repeats. 

Solution: use Cloud feature to pin replica in failure scenario so both operations are sent to same replica. 
This solution is not a silver bullet and there is still chance to failure because problem is actually classic of distributed systems that is hard to solve with a single mechanism

chernser:
## Summary

See https://github.com/ClickHouse/clickhouse-kafka-connect/issues/827 for problem description

- Adds property `enableReplicaPinning` . Affects error handling when schema mismatch 
- Adds sending `X-ClickHouse-Replica-Tag` when pinning enabled and error detected. 

Closes: https://github.com/ClickHouse/clickhouse-kafka-connect/issues/827

## Checklist

### DIFF
--- a/src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
+++ b/src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
@@ -997,8 +997,13 @@ public class ClickHouseWriter implements DBWriter {
             Optional<String> updateTableException = UPDATE_TABLE_EXCEPTION_STR_TO_ERROR_CODE.keySet().stream().filter(code -> e.getMessage().contains(code)).findFirst();
             if (e.getMessage() != null && updateTableException.isPresent() && retry) {
                 LOGGER.warn("Error code {}. Trying to update table mapping because ClickHouse table schema may have evolved.", UPDATE_TABLE_EXCEPTION_STR_TO_ERROR_CODE.get(updateTableException.get()));
-                Table tableTmp = urgentTableUpdate(table);
-                doInsertRawBinary(records, tableTmp, queryId, tableTmp.hasDefaults(), false);
+                try {
+                    chc.pinReplica();
+                    Table tableTmp = urgentTableUpdate(table);
+                    doInsertRawBinary(records, tableTmp, queryId, tableTmp.hasDefaults(), false);
+                } finally {
+                    chc.unpinReplica();
+                }
             } else {
                 LOGGER.error("Error inserting records", e);
                 throw e;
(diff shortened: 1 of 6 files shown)


src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
```diff
@@ -558,6 +578,81 @@ public List<Table> extractTablesMapping(String database, Map<String, Table> cach
...
+                }
+                request.option(ClickHouseHttpOption.CUSTOM_HEADERS, customHeaders);
+            } catch (Exception e) {
+                LOGGER.error("Failed to set replica tag header", e);
+            }
+        }
+        return request;
+    }
+
+    public void setReplicaTagHeaderV2(QuerySettings settings) {
```
mzitnik, line 634: Looking into the Java client repo, maybe we can inherit from CommonSettings and here to use CommonSettings instead of having 2 different methods

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
```diff
@@ -1049,6 +1060,8 @@ protected void doInsertRawBinaryV2(List<Record> records, Table table, QueryIdent
 
         InsertSettings insertSettings = new InsertSettings();
         insertSettings.setDatabase(database);
+        chc.setReplicaTagHeaderV2(insertSettings);
+        chc.setReplicaTagHeaderV2(insertSettings);
```
mzitnik, line 1064: seems redundant

mzitnik (changes requested):

mzitnik:
@chernser do you think we are able to run a test without mock and simulate the real issue?

src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java
chernser: `CommonSettings` are aggregated inside `InsertSettings` and `QuerySettings` then it will require extra copy and more code changes.

src/main/java/com/clickhouse/kafka/connect/sink/db/ClickHouseWriter.java
chernser: will remove.

chernser:
@mzitnik 
Thank you for the review! 

This feature is only Cloud one and implementation is on LB side. I've created issue https://github.com/ClickHouse/clickhouse-kafka-connect/issues/831
Currently Mock Server at least verifies that proper header is set.

mzitnik:
@chernser, we have one failing test here

mzitnik (approved):
but lets not document this new flag until we will have testing with cloud and not mock

chernser:
@mzitnik 
got it!

Issue #528, opened 2025-05-06
Expose in configuration client compression


mzitnik:
**Is your feature request related to a problem? Please describe.**
Currently, we are not exposing client compression; by default, it is disabled. 

**Describe the solution you'd like**
Add to the configuration and set the client compression to true by default.

PR #830, opened 2026-09-07
Expose client request compression as a connector config for v2

pcalcado:
Implements [528 - Expose in configuration client compression](https://github.com/clickhouse/clickhouse-kafka-connect/issues/528) for the V2 client.

This exposes client request compression as a connector setting. The connector now accepts a single boolean `clientCompression` flag. The default is `false`, preserving existing behavior unless users opt in.

When enabled, the connector applies the client library's native request-compression toggle on the V2 insert path and on V2 exactly-once state-store writes. To avoid surprising behavior on the older client implementation, `clientCompression=true` is rejected when the connector would use V1, including the current unset `client_version` default.

The scope is intentionally narrow: this PR exposes the connector-level boolean without expanding the configuration surface around compression algorithms.

Tests cover the connector/config seams. Because test coverage is not good enough for this aprticular feature, I first considered adding it to the benchmarks but thought this might be just re-testing the client and it would differ from what exists in the repo already. Instead, I built a separate lab just to manually validate that compression happens as expected. It's in a branch on ky fork:

- https://github.com/pcalcado/clickhouse-kafka-connect/pull/2

In that lab, V2 JSON inserts of a 100k-row mixed-cardinality profile payload showed median captured wire bytes drop from ~111MB to ~95MB with `clientCompression=true`. Elapsed time was slightly higher in that run, so I am treating this as wire-behavior evidence of the PR working. Happy to do something else to validate if this isn't good enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * New `retryOnSocketException` setting (default `false`). When enabled, a `java.net.SocketException` (broken pipe,
   connection reset) from the ClickHouse client is retried like the existing timeout cases instead of failing the
   task. With client V2's persistent connections a single server-side connection reset otherwise fails every task at once.
-* Added `clientCompression` connector configuration to control ClickHouse client request compression for the V2 insert and exactly-once state-store paths. The default is `false`, preserving existing behavior unless users opt in. Configuring `clientCompression=true` with `clientVersion=V1` is rejected. (https://github.com/ClickHouse/clickhouse-kafka-connect/issues/528)
+* Added `clientCompression` connector configuration to control ClickHouse client request compression for V2 insert traffic and exactly-once state-store writes. The default is `false`, preserving existing behavior unless users opt in. Configuring `clientCompression=true` with `client_version=V1`, including the unset `client_version` default, is rejected. (https://github.com/ClickHouse/clickhouse-kafka-connect/issues/528)
 
 ## Bug Fixes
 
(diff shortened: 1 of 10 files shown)


chernser:
@pcalcado 

However I have a question: do we need a separate compression flag in connector? 
Usually we pass client configuration via `jdbcConnectionProperties`

pcalcado:
> @pcalcado
> 
> Thank you for the contribution!
> 
> However I have a question: do we need a separate compression flag in connector? Usually we pass client configuration via `jdbcConnectionProperties`

hey @chernser thanks for taking a look

happy to make it use `jdbcConnectionProperties`, if i understand correctly it's currently not passed at all and I have to use `clientBuilder.setOptions(optsAsMap)` does this make sense?

we've been using this for v1,m but i had the impression that for V2 the preferred way was to use the typed client API, e.g. `Client.Builder.compressClientRequest(...)` / `InsertSettings.compressClientRequest(...)`, rather than relying on URL query parameters--i guess i misinterpreted that?

chernser:
@pcalcado 
Previously there was a bug that v2 client has not parsed it and only clickhouse own settings were used. 
Now it is fixed - I'm looking at https://github.com/ClickHouse/clickhouse-kafka-connect/blob/eb4eb3c67c240f607fcd3370d849c092840eaed2/src/main/java/com/clickhouse/kafka/connect/sink/db/helper/ClickHouseHelperClient.java#L173

I will appreciate if you can confirm it works for you.

Thank you!

pcalcado:
oh @chernser i guess i had an older version and misinterpreted the design then. ill close this and report back if there's anything wonky

thank you

PR #837, opened 2026-09-22
FieldToJsonString transformation that serializes one or more (optiona…

Dmitry-Semashko:
FieldToJsonString Transformation
This transformation serializes one or more record fields into their JSON string representation.
It is useful when a source message contains a nested object (or array) but the destination column
is a plain `String` - for example when you want to keep a free-form, schema-less payload in a single
ClickHouse `String` column instead of modelling it as a nested `Tuple`/`Map`.


Dmitry-Semashko:
closing as not needed this SMT for clickhouse sink. It can be implemented for any kafka connect pipeline

